PR #31, opened 2021-09-17
fix(rest): routing fix and moved cors to dependencies

janrtvld:
fix(rest): routing fix and moved cors to dependencies


TimoGlastra (approved):

packages/rest/README.md
```diff
@@ -43,3 +43,27 @@ Then add the rest package to your project.
 ```sh
 yarn add @aries-framework/rest
 ```
+
+### Quick start
+
+> The API is generated using the OpenAPI Schema (Swagger). However this schema is not representing the real API. A lot of types are not correct. Keep this in mind when using this package.
```
TimoGlastra, line 49: The API isn't generated, but the OpenAPI spec is. I think it would be good to make it clear that the API isn't generated.

PR #38, opened 2021-10-01
feat(react-hooks): added react-hooks library

NeilSMyers:
Signed-off-by: NeilSMyers <mmyersneil@gmail.com>


packages/aries-hooks/README.md
```diff
@@ -0,0 +1,68 @@
+# Aries Hooks
```
blu3beri, line 1: I think the package name could be just `@aries-framework/hooks`.

packages/aries-hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+This package exposes useful hooks that allow you to easily interact with AFJ.
+
+Everything exported from Aries Hooks:
+
+```
```
blu3beri, line 7: use 
"```ts" for syntax highlighting

packages/aries-hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+	useProofs,
+	useProofById,
+	useProofByState
+} from 'aries-hooks'
+```
+
+First step is to wrap your entire app in our `<AgentProvider/>`. The provider takes two props, the first is your `agentConfig` object and the second is your `genesisUrl`. The base of your app should look something like this:
+
+```
+import AgentProvider from "aries-hooks"
```
blu3beri, line 26:
```suggestion
import { AgentProvider } from "@aries-framework/hooks"
```

packages/aries-hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+const App = () => {
+	return (
+		<AgentProvider agentConfig={Your full agent config object} genesisUrl={Your genesis url} >
```
blu3beri, line 30: Do we want to do the genesis via an url? I am not a 100% sure it is a must feature. It could always be handled like this by the framework consumer.

packages/aries-hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+	return (
+		<Flatlist
+			data={connections}
+			renderItem={({item}) => <MyListItem connection={item} />}
+		/>
+	)
+}
+```
+
+The three base hooks: `useConnections`, `useCredentials`, and `useProofs` work just like the above! Just call the hook, destructure the data, and pass it through!
```
blu3beri, line 59: My personal preference but I think it would be better without the exclamantion marks.

packages/aries-hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Each base hook has a `ById` version that returns a singular record. For example if I wanted only a specific connectionRecord, I'd do this.
+`const { connection } = useConnection(id)`
+
+More commonly, you'll want to get a filtered list of records based off of their state. Well, Hooray! We have a `ByState` version as well. For example, you can do this:
+`const credentials = useCredentialByState(CredentialState.OfferReceived)`
+
+Boom Bam Baby!
+That's all you need to know to get that data flowinggg.
```
blu3beri, line 68: While the data is indeed flowing I think we should not include these two sentences ;).

packages/aries-hooks/genesis-utils.ts
```diff
@@ -0,0 +1,18 @@
+import axios from 'axios'
+import RNFS from 'react-native-fs'
+
+async function storeGenesis(genesis: string, fileName: string): Promise<string> {
```
blu3beri, line 4: Again, I think this could be something a framework consumer would implement.

packages/aries-hooks/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@aries-framework/aries-hooks",
```
blu3beri, line 2: I prefer the name `@aries-framework/hooks`.

packages/aries-hooks/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "version": "1.0.1",
+  "description": "",
```
blu3beri, line 4: Could you supply a description for the package?

packages/aries-hooks/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@aries-framework/aries-hooks",
+  "version": "1.0.1",
```
blu3beri, line 3: Version would start at `0.0.1`.

packages/aries-hooks/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "main": "index.tsx",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "@aries-framework/core": "0.1.0-alpha.273",
+    "@aries-framework/react-native": "0.1.0-alpha.273",
+    "indy-sdk-react-native": "^0.1.14"
```
blu3beri, line 14: Is this necessary?

packages/aries-hooks/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@aries-framework/aries-hooks",
+  "version": "1.0.1",
+  "description": "",
+  "main": "index.tsx",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
```
blu3beri, line 9: remove field.

packages/aries-hooks/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@aries-framework/aries-hooks",
+  "version": "1.0.1",
+  "description": "",
+  "main": "index.tsx",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
blu3beri, line 7: Maybe in a seperate PR, but some small tests would be nice.

packages/aries-hooks/src/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+  BasicMessageRecord,
+  BasicMessageReceivedEvent,
+  BasicMessageEventTypes,
+} from '@aries-framework/core'
+
+import { agentDependencies } from '@aries-framework/react-native'
+
+const TAG = 'Aries-Hooks:'
+
+const AgentContext = createContext<any>({})
```
blu3beri, line 31: Could you replace every `any` type with the correct typing?

packages/aries-hooks/src/index.tsx
```diff
@@ -0,0 +1,295 @@
+import React, { createContext, useState, useEffect, useContext } from 'react'
```
blu3beri, line 1: I think this files extension could just be `.ts`, right?

packages/aries-hooks/src/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+const TAG = 'Aries-Hooks:'
+
+const AgentContext = createContext<any>({})
+const ConnectionContext = createContext<any>({})
+const CredentialContext = createContext<any>({})
+const ProofContext = createContext<any>({})
+const BasicMessageContext = createContext<any>({})
+
+// Agent
```
blu3beri, line 37: I think it would be nice to split every set of hooks into their on file.

packages/aries-hooks/src/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+  CredentialRecord,
+  WsOutboundTransport,
+  BasicMessageRecord,
+  BasicMessageReceivedEvent,
+  BasicMessageEventTypes,
+} from '@aries-framework/core'
+
+import { agentDependencies } from '@aries-framework/react-native'
+
+const TAG = 'Aries-Hooks:'
```
blu3beri, line 29: I do like the idea of prefixing the logs with some indicator like this. However I am not sure if this is the way to go. Would like to hear other opinions about this.

packages/aries-hooks/src/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+// Connection
+export const useConnections = (): { connections: ConnectionRecord[]; loading: boolean } => {
+  return useContext(ConnectionContext)
+}
+
+export const useConnectionById = (id: string): ConnectionRecord | undefined => {
+  const { connections } = useContext(ConnectionContext)
+  const connection = connections.find((c: ConnectionRecord) => c.id === id)
+  if (!connection) {
+    console.warn(TAG, `Could not find Connection By ID ${id}`)
```
blu3beri, line 51: No `console.xxx` please. Maybe we could use the agent logger or just return a null.

packages/aries-hooks/src/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+const AgentProvider: React.FC<Props> = ({ agentConfig, genesisUrl, children }) => {
+  const [agentState, setAgentState] = useState<{
+    agent: Agent | null
+    loading: boolean
+  }>({
+    agent: null,
+    loading: true,
+  })
+  const [connectionState, setConnectionState] = useState<{ connections: ConnectionRecord[] | []; loading: boolean }>({
```
blu3beri, line 123:
```suggestion
  const [connectionState, setConnectionState] = useState<{ connections: ConnectionRecord[]; loading: boolean }>({
```

blu3beri (changes requested):
I like the addition of the hooks and could definitely be expanded upon in the future. I left some initial comments. I don't think that these are the only points but it should get it in the right direction.

Thanks for the contribution!

packages/hooks/README.md
```diff
@@ -0,0 +1,68 @@
+# Aries Framework Hooks
+
+This package exposes useful hooks that allow you to easily interact with AFJ.
+
+Everything exported from Hooks:
```
TimoGlastra, line 5: I think we should be more clear that this is about react hooks.

```suggestion
# React Hooks for Aries Framework JavaScript

This package exposes useful React hooks that allow you to easily interact with AFJ.

Everything exported from Hooks:
```

packages/hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+	useConnections,
+	useConnectionById,
+	useConnectionByState,
+	useCredentials,
+	useCredentialById,
+	useCredentialByState,
+	useProofs,
+	useProofById,
+	useProofByState
+} from '@aries-framework/hooks'
```
TimoGlastra, line 20: I'd also call it `@aries-framework/react-hooks` to make it extra clear.

packages/hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+First step is to wrap your entire app in our `<AgentProvider/>`. The provider takes two props, the first is your `agentConfig` object and the second is your `genesisUrl`. The base of your app should look something like this:
```
TimoGlastra, line 23: As @blu3beri already pointed out, I don't think we should promote the usage of `genesisUrl` as it will download the genesis every time you start the agent. If this is really a desired feature, agents can add this functionality themselves

packages/hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+```ts
+import AgentProvider from "@aries-framework/hooks"
+
+const App = () => {
+	return (
+		<AgentProvider agentConfig={Your full agent config object} genesisUrl={Your genesis url} >
```
TimoGlastra, line 30: Also, I'd rather have the `AgentProvider` take an `agent` instance as input, this gives a lot more flexibility to setup the agent to your requirements, while still leveraging this package

packages/hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+			// Your app here
+		</AgentProvider>
+	)
+}
+```
+
+And that's it! Your app should be set up to recieve all the necessary data your app will need! Now let's see how we actually get that data to our components.
+
+The `useAgent` hook returns `{ agent, loading }` so anytime you need access to any of the methods tied to the agent, you can `useAgent()` anywhere.
```
TimoGlastra, line 39: I prefer to use `isLoading` over `loading` as it feels more explicit

```suggestion
The `useAgent` hook returns `{ agent, isLoading }` so anytime you need access to any of the methods tied to the agent, you can `useAgent()` anywhere.
```

packages/hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+The following is an example of how you could use the `useConnections` hook to render a full list of all a user's connections.
+
+```ts
+import { useConnections } from 'aries-hooks'
```
TimoGlastra, line 44: outdated package import

packages/hooks/README.md
```diff
@@ -0,0 +1,68 @@
...
+			renderItem={({item}) => <MyListItem connection={item} />}
+		/>
+	)
+}
+```
+
+The three base hooks: `useConnections`, `useCredentials`, and `useProofs` work just like the above! Just call the hook, destructure the data, and pass it through!
+
+Each base hook has a `ById` version that returns a singular record. For example if I wanted only a specific connectionRecord, I'd do this.
+`const { connection } = useConnection(id)`
```
TimoGlastra, line 62: Maybe we could also return the connection record directly, instead of as part of an object, or are there also other properties returned?

```suggestion
`const { connection } = useConnectionById(id)`
```

packages/hooks/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@aries-framework/hooks",
```
TimoGlastra, line 2: Please copy the package.json (from e.g. redux-store) as closely as possible. It should at least contain the `publishConfig`, `scripts`, main and build files. (and probably some other properties)

packages/hooks/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "This package exposes useful hooks that allow you to easily interact with AFJ.",
+  "main": "index.tsx",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "@aries-framework/core": "0.1.0-alpha.273",
+    "@aries-framework/react-native": "0.1.0-alpha.273",
```
TimoGlastra, line 13: Could you update it to the same as other packages in this repo (276)?

Also, peerDependencies should be listed in devDependencies (see redux-store as example). Also, I don't think it is strictly required to have the react native package in here, it's only using the core package

packages/hooks/jest.config.ts
```diff
@@ -0,0 +1,13 @@
+import type { Config } from '@jest/types'
+
+import base from '../../jest.config.base'
+
+import packageJson from './package.json'
+
+const config: Config.InitialOptions = {
```
TimoGlastra, line 7: Could you at a placeholder for tests, I think this will fail without any tests. See here how to add a todo test: https://github.com/hyperledger/aries-framework-javascript-ext/blob/main/packages/redux-store/tests/index.test.ts

TimoGlastra (changes requested):
Nice work @NeilSMyers! I've left some comments

packages/hooks/README.md
JamesKEbert: Could this be a promise of an agent? The issue I think I saw here is that we need to have the context providers be initialized at run time.

TimoGlastra: I guess that would work. Is it because you are not in an async context when passing the agent to the agent provider?

packages/hooks/README.md
TimoGlastra: Rather not. It's not part of the feature of this library so we shouldn't include IMO.

packages/react-hooks/README.md
```diff
@@ -0,0 +1,72 @@
+# React Hooks for Aries Framework JavaScript
+
+This package exposes useful React hooks that allow you to easily interact with AFJ.
+
+Everything exported from Hooks:
+
+```ts
```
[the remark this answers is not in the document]
JamesKEbert: @NeilSMyers would you be able to look at doing so?

packages/hooks/README.md
JamesKEbert: Can do. 👍 @NeilSMyers could you add this as an item to take care of come Monday?

packages/hooks/README.md
JamesKEbert: I actually realized at like 3 in the morning last night that this was a somewhat dumb POV of mine--we've got built into the hooks package state properties for loading state--which means we just have to change the AgentProvider to be able to handle having `Agent` being empty initially while it initializes asynchronously outside the AgentProvider component. 
There might be some additional nuances to consider there, but overall I'd agree with this. @NeilSMyers we should also look at this.

packages/hooks/README.md
NeilSMyers: so am I removing the genesisUrl from everything?

packages/react-hooks/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@aries-framework/react-hooks",
+  "main": "build/index",
+  "types": "build/index",
+  "version": "0.0.1",
```
TimoGlastra, line 5: "version": "0.0.0"

packages/react-hooks/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "prepublishOnly": "yarn run build",
+    "test": "jest"
+  },
+  "devDependencies": {
+    "@aries-framework/core": "0.1.0-alpha.276",
+    "react": "^17.0.2",
+    "typescript": "~4.4.2"
+  },
+  "peerDependencies": {
+    "@aries-framework/core": "0.1.0-alpha.276"
```
TimoGlastra, line 32: react should be a peer dependency

packages/react-hooks/src/index.tsx
```diff
@@ -0,0 +1,277 @@
...
+// Connection
+export const useConnections = (): { connections: ConnectionRecord[]; loading: boolean } => {
+  return useContext(ConnectionContext)
+}
+
+export const useConnectionById = (id: string): ConnectionRecord | undefined => {
+  const { connections } = useContext(ConnectionContext)
+  const connection = connections.find((c: ConnectionRecord) => c.id === id)
+  if (!connection) {
+    console.warn(TAG, `Could not find Connection By ID ${id}`)
```
TimoGlastra, line 44: please remove logging or use the agent logger

packages/react-hooks/src/index.tsx
```diff
@@ -0,0 +1,277 @@
...
+        setCredentialState({ credentials, loading: false })
+        setProofState({ proofs, loading: false })
+        setBasicMessageState({ basicMessages, loading: false })
+      }
+    } catch (e) {
+      console.error(TAG, 'ERROR IN SET_INITIAL_STATE:', e)
+    }
+  }
+
+  useEffect(() => {
```
TimoGlastra, line 155: fine to merge now, but I think it would be nice to split into a separate file per feature (connections, credentials, proofs, etc..)

packages/react-hooks/src/index.tsx
```diff
@@ -0,0 +1,277 @@
...
+          credentials: newCredentialsState,
+        })
+      }
+
+      agentState.agent?.events.on(CredentialEventTypes.CredentialStateChanged, listener)
+
+      return () => {
+        agentState.agent?.events.off(CredentialEventTypes.CredentialStateChanged, listener)
+      }
+    }
```
TimoGlastra, line 205: why are all listeners removed during loading? This means if an event is fired during loading we won't get it in our local state

packages/react-hooks/src/index.tsx
JamesKEbert: I think this does provide some valuable info--not sure if it necessarily belongs in the AFJ logger, since this is an extension? Not super opinionated here though. 
On a second note, can we actually use the logger outside of AFJ? I wasn't aware that was being exposed properly (or maybe I'm being dumb 😆 )

packages/react-hooks/package.json
JamesKEbert: @NeilSMyers

packages/react-hooks/src/index.tsx
JamesKEbert: > Why do we not add the loaders when loading=true?

Since we're loading the `agent` asynchronously (which differs from the redux-store package _I think_), then there's a period of time in the aries-hooks package that `agent` is undefined, so if we try and register the listener while loading = `true` then we'll see an error trying to register the listener using something undefined.
AFAIK the cleanup function isn't triggered immediately, only when the useEffect is being ran again or the component is being unmounted. I could be wrong in that statement though.

I am all ears if I am missing something stupid here! :)

JamesKEbert: No worries--totally could see your concern if you didn't see how the loading param was working.

packages/react-hooks/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@aries-framework/react-hooks",
+  "main": "build/index",
+  "types": "build/index",
+  "version": "0.0.0",
```
blu3beri, line 5: Shouldn't this be 0.1.0 for the initial version

blu3beri (approved):

PR #41, opened 2021-10-08
feat(push-notifications): added push-notifications plugin

blu3beri:
- Push notifications module that can be injected into the agent to set/get a push notification configuration
- only the notification-receiver side
- Includes tests and a sample

Signed-off-by: Berend Sliedrecht <berend@animo.id>


packages/push-notifications/src/services/PushNotificationsService.ts
```diff
@@ -0,0 +1,31 @@
...
+@scoped(Lifecycle.ContainerScoped)
+export class PushNotificationsService {
+  public constructor(private connectionService: ConnectionService) {}
+
+  public async processSetDeviceInfo(deviceInfo: DeviceInfo, connectionId: string) {
+    const message = new PushNotificationsSetDeviceInfoMessage(deviceInfo)
+    const connection = await this.connectionService.getById(connectionId)
+    return { message, connection }
+  }
+
```
TimoGlastra, line 25: The convention we have for services is 

`processXXX` for processing a received message and `createXXX` for creating a message to send. So in this case I'd use `createSetDeviceInfo`

packages/push-notifications/src/messages/PushNotificationsSetDeviceInfoMessage.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    if (options) {
+      this.deviceToken = options.deviceToken
+      this.deviceVendor = options.deviceVendor
+    }
+  }
+
+  @Equals(PushNotificationsSetDeviceInfoMessage.type)
+  public readonly type = PushNotificationsSetDeviceInfoMessage.type
+  public static readonly type = 'https://didcomm.org/push-notifications/1.0/set-device-info'
```
TimoGlastra, line 24: This doesn't have the different messages for services yet right? (set-native-device-info, set-fcm-device-info, set-expo-device-info)

packages/push-notifications/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+In order for this plugin to work we have to inject it into the agent to access agent functionality. See the example for more information
+
+### Example of usage
+
+> This is the current way however this will be changed someday to improve plugin management
+
+```ts
+// Agent already has been setup BUT NOT YET INITIALIZED
+import { PushNotificationsModule } from '@aries-framework/push-notifications'
```
TimoGlastra, line 57: I'd add an agent = new Agent({ /** agent config .. */ }) line to make it more clear you create the agent before calling resolve

packages/push-notifications/src/messages/PushNotificationsGetDeviceInfoMessage.ts
```diff
@@ -0,0 +1,13 @@
+import { AgentMessage } from '@aries-framework/core/build/agent/AgentMessage'
+import { Equals } from 'class-validator'
+
+/**
+ * Message to get the device information from another agent for push notifications
+ *
+ * @todo ADD RFC
+ */
+export class PushNotificationsGetDeviceInfoMessage extends AgentMessage {
```
TimoGlastra, line 9: This is missing the device info message: https://didcomm.org/push-notifications/1.0/device-info  as response to get-device-info. As that is a message you receive as client we need to add the message and also a handler for it

packages/push-notifications/tests/utils/logger.ts
```diff
@@ -0,0 +1,87 @@
+import type { ILogObject } from 'tslog'
```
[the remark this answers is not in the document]
blu3beri: Agreed.

packages/push-notifications/src/PushNotificationsModule.ts
```diff
@@ -0,0 +1,66 @@
+import type { DeviceInfo } from './services'
+
+import { Dispatcher } from '@aries-framework/core/build/agent/Dispatcher'
+import { MessageSender } from '@aries-framework/core/build/agent/MessageSender'
+import { createOutboundMessage } from '@aries-framework/core/build/agent/helpers'
```
TimoGlastra, line 5: I think we should also export `Dispatcher` and `MessageSender` from the framework (not so sure on the createOutboundMessage)

packages/push-notifications/src/messages/PushNotificationsDeviceInfoMessage.ts
```diff
@@ -0,0 +1,46 @@
...
+  @Expose({ name: 'device_token' })
+  @IsString()
+  public deviceToken!: string
+
+  @Expose({ name: 'device_vendor' })
+  @IsString()
+  public deviceVendor!: DeviceVendor
+
+  @Expose({ name: 'service' })
+  @IsEnum(String)
```
TimoGlastra, line 44: How does this work?

I think it should just be string right cause you didn't define an `enum` type

```suggestion
  @IsEnum(String)
```

packages/push-notifications/src/services/PushNotificationsService.ts
```diff
@@ -0,0 +1,45 @@
...
+@scoped(Lifecycle.ContainerScoped)
+export class PushNotificationsService {
+  public constructor(private connectionService: ConnectionService) {}
+
+  public async createSetNativeDeviceInfo(connectionId: string, deviceInfo: DeviceInfo) {
+    const message = new PushNotificationsSetNativeDeviceInfoMessage(deviceInfo)
+    const connection = await this.connectionService.getById(connectionId)
+    return { message, connection }
+  }
```
TimoGlastra, line 26: Sorry for not mentioning last review but we usually use ids on the module layer and records on the service layer. Then you also don't need to return the connection record, as the module already has it. Then we should probably also assert the connection is ready 

```suggestion
  public async createSetNativeDeviceInfo(connectionRecord: ConnectionRecord, deviceInfo: DeviceInfo) {
    return new PushNotificationsSetNativeDeviceInfoMessage(deviceInfo)
  }
```

e.g. this is from the discover features module:

```
  public async queryFeatures(connectionId: string, options: { query: string; comment?: string }) {
    const connection = await this.connectionService.getById(connectionId)
    
    // This is not in the discover features module, but we should check if the connection is ready
    connection.assertReady()

    const queryMessage = await this.discoverFeaturesService.createQuery(options)

    const outbound = createOutboundMessage(connection, queryMessage)
    await this.messageSender.sendMessage(outbound)
  }
```

TimoGlastra: Updated because we don't even need the connection record in the service now anymore

packages/push-notifications/src/PushNotificationsModule.ts
blu3beri: Ive added `Dispatcher` and `MessageSender` to the [pr](https://github.com/hyperledger/aries-framework-javascript/pull/480).

packages/push-notifications/src/PushNotificationsModule.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  /**
+   * Sends the native device info (token and vendor) to another agent via a `connectionId`
+   */
+  public async sendNativeDeviceInfo(connectionId: string, deviceInfo: DeviceInfo) {
+    const connection = await this.connectionService.getById(connectionId)
+    const message = this.pushNotificationService.createSetNativeDeviceInfo(deviceInfo)
+
+    connection.assertReady()
```
TimoGlastra, line 33: I'd swap these around. The service doesn't store anything so it doesn't matter in this case, but if the connection is not ready we don't have to call the service

packages/push-notifications/src/messages/PushNotificationsDeviceInfoMessage.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (options) {
+      this.deviceToken = options.deviceToken
+      this.deviceVendor = options.deviceVendor
+      this.notificationsService = options.service
+    }
+  }
+
+  @Equals(PushNotificationsDeviceInfoMessage.type)
+  public readonly type = PushNotificationsDeviceInfoMessage.type
+  public static readonly type = 'https://didcomm.org/push-notifications/1.0/device-info'
```
TimoGlastra, line 33: incorrect type

packages/push-notifications/src/messages/PushNotificationsGetDeviceInfoMessage.ts
```diff
@@ -0,0 +1,13 @@
...
+  @Equals(PushNotificationsGetDeviceInfoMessage.type)
+  public readonly type = PushNotificationsGetDeviceInfoMessage.type
+  public static readonly type = 'https://didcomm.org/push-notifications/1.0/get-device-info'
```
TimoGlastra, line 12: incorrect type

packages/push-notifications/tests/utils/logger.ts
```diff
@@ -25,6 +25,8 @@ export class TestLogger extends BaseLogger {
   public constructor(logLevel: LogLevel, name?: string) {
     super(logLevel)
 
+    this.logLevel.toExponential()
```
TimoGlastra, line 28: ?

PR #58, opened 2021-10-28
fix(react-hooks): message filtering overflows

NeilSMyers:
added a use effect to useBasicMessagesByConnectionId because the filter was causing too many renders in use. This is being created at the same time as a new AFJ event listener for sending messages and will rely on that event to work


JamesKEbert (approved):
LGTM! This may also want to wait to be merged after decisions on AFJ #https://github.com/hyperledger/aries-framework-javascript/pull/507 are finalized.

TimoGlastra (approved):
LGTM

PR #92, opened 2022-02-03
feat: webhooks

### LINKED ISSUE #63: [rest] Add web hooks and events to rest server
Currently you can interact with the rest API by making HTTP requests. However there is no way for the rest API to call back to the controller in case of certain events. This means you need to use polling. 

We should add  a call-back method so the rest api can call the controller.

The current method we're thinking of is Webhooks. Whenever an event is triggered in the Agent, the rest wrapper will send the event to an HTTP endpoint as configured in the rest wrapper

1. Determine all event types we want to listen to in Aries Framework JavaScript. We currently support connection, credential and proof types in the rest api wrapper.
2. Add a webhookUrl option to the rest `setupServer` method that specifies where the webhooks should be sent to
3. Listen to the event types and whenever a call is made, send a webhook event to the registered webhookUrl

`baseWebhookUrl` -> `https://example.com/webhooks` -> `https://example.com/webhooks/{topic}` -> `https://example.com/webhooks/connections` (`connections`, `proofs`, `credentials`)

- Whenever an event is received for one of the topics, make a POST request to the webhook endpoint with the event payload.

```
POST https://example.com/webhooks/connections
{
   "id": "1234-1234-1234-1234"
}
```

```ts
agent.events.on(ConnectionEventTypes.ConnectionStateChanged, (event) => {

  const record =  event.payload.connetionRecord
  const body = record.toJSON()

  // use either library or native Node.JS (node-fetch)
  await post("https://example.com/webhooks/connections", body)
})
```

Annelein:
Fixes #63


packages/rest/src/events/ConnectionEvents.ts
```diff
@@ -0,0 +1,17 @@
...
+export const connectionEvents = async (agent: Agent, webhookUrl: string) => {
+  agent.events.on(ConnectionEventTypes.ConnectionStateChanged, async ({ payload }: ConnectionStateChangedEvent) => {
+    const record = payload.connectionRecord
+    const body = record.toJSON()
+
+    await fetch(webhookUrl + '/connections', {
+      method: 'POST',
+      body: JSON.stringify(body),
+      headers: { 'Content-Type': 'application/json' },
+    })
```
TimoGlastra, line 15: I think it would be good to extract the functionality into a separate method so it can be used by all the event handlers, this makes it easier to keep them consistent.

packages/rest/src/index.ts
```diff
@@ -5,8 +5,8 @@ import type { Agent } from '@aries-framework/core'
 
 import { setupServer } from './server'
 
-export const startServer = async (agent: Agent, config: ServerConfig) => {
-  const server = await setupServer(agent, config)
+export const startServer = async (agent: Agent, config: ServerConfig, webhookUrl?: string) => {
+  const server = await setupServer(agent, config, webhookUrl)
```
TimoGlastra, line 9: Maybe we can add the webhookUrl to the `ServerConfig`?

packages/rest/src/server.ts
```diff
@@ -9,10 +9,14 @@ import { routingControllersToSpec } from 'routing-controllers-openapi'
...
 
+import { connectionEvents } from './events/ConnectionEvents'
+import { credentialEvents } from './events/CredentialEvents'
+import { proofEvents } from './events/ProofEvents'
+
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const packageJson = require('../package.json')
 
-export const setupServer = async (agent: Agent, config: ServerConfig) => {
+export const setupServer = async (agent: Agent, config: ServerConfig, webhookUrl?: string) => {
```
TimoGlastra, line 19: Same here

package-lock.json
```diff
@@ -0,0 +1,18280 @@
+{
```
TimoGlastra: We use yarn instead of NPM. Can you remove this file?

TimoGlastra (commented):
Nice @Annelein

Would be good to add some tests before merging, but otherwise looks good!

TimoGlastra:
Also, seems like tests are failing due to node-fetch. I'd recommend to downgrade the node-fetch version to the latest 2.x version. They made some breaking changes in v3 that we aren't ready to consume et

packages/rest/package.json
```diff
@@ -35,7 +35,8 @@
     "routing-controllers": "^0.9.0",
     "routing-controllers-openapi": "^3.1.0",
     "swagger-ui-express": "^4.1.6",
-    "typedi": "^0.10.0"
+    "typedi": "^0.10.0",
+    "node-fetch": "2.2.0"
```
TimoGlastra, line 39: Er is al en hogere versie beschikbaar:

```suggestion
    "node-fetch": "^2.6.7"
```

packages/rest/src/events/ConnectionEvents.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { ConnectionEventTypes } from '@aries-framework/core'
+
+import { webhookEvent } from './WebhookEvent'
+
+export const connectionEvents = async (agent: Agent, config: ServerConfig) => {
+  agent.events.on(ConnectionEventTypes.ConnectionStateChanged, async ({ payload }: ConnectionStateChangedEvent) => {
+    const record = payload.connectionRecord
+    const body = record.toJSON()
+    webhookEvent(config.webhookUrl + '/connections', body)
```
TimoGlastra, line 12: Do we need to await the webhookEvent method?

TimoGlastra: Nit, but maybe `sendWebhookEvent` makes the intent of the function more clear?

packages/rest/tests/webhook.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  afterEach(async () => {
+    await agent.shutdown()
+    await agent.wallet.delete()
+    server.close()
+  })
+
+  test('should return a webhook event when connection state changed', async () => {
+    const { connectionRecord } = await agent.connections.createConnection()
+
+    const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
```
TimoGlastra, line 40: Maybe we can define this once in the `utils/helpers` file and reuse for all tests? Declaring it multiple times seems uncessary

packages/rest/tests/webhook.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    await sleep(100)
```
TimoGlastra, line 41: Also: Maybe good to add a comment why we're using this sleep, so it's clear for future developers

packages/rest/README.md
```diff
@@ -64,3 +64,15 @@ const run = async (agent: Agent) => {
...
+### Webhooks
+
+We use webhooks as a method for the rest api to have the option to call the controller in case of an event.
+
+Usage: add a webhook url to the ServerConfig you pass to startServer()
+
+Current supported events are:
+
+- connections
+- credentials
```
TimoGlastra, line 77: Could you elaborate a bit on the usage?

- How do you pass the `webhookUrl` (can be passed to both `startServer` and `setupServer`) -- Code examples are always nice!
- How is the `webhookUrl` used? Does it send everything to the webhook url, or does it append the topic to the url before sending? This is useful for people integrating with it.
- What is the payload sent in the body of the webhook? Is it a json object? If so, what is the content of the payload?

TimoGlastra: E.g. look at the ACA-Py docs as an example: https://github.com/hyperledger/aries-cloudagent-python/blob/main/AdminAPI.md#administration-api-webhooks

I don't think we should add documentation on a per parameter level (as that becomes quickly outdated), but would be good to add some more context

packages/rest/README.md
```diff
@@ -64,3 +64,34 @@ const run = async (agent: Agent) => {
...
+  const config = {
+    port: 3000,
+    webhookUrl: 'http://test.com',
+  }
+  await startServer(agent, config)
+}
+run()
+```
+
+In case of an event, we will sent the event to the webhookUrl with the topic of the event added to the url (http://test.com/{topic}).
```
TimoGlastra, line 93:
```suggestion
In case of an event, we will send the event to the webhookUrl with the topic of the event added to the url (http://test.com/{topic}).
```

packages/rest/README.md
```diff
@@ -64,3 +64,34 @@ const run = async (agent: Agent) => {
...
+
+So in this case when a connection event is triggered, it will be sent to: http://test.com/connections
+
+The payload of the webhook contains the date, the header, topic and a body containing the connection_id, the state of the event and possibly a key, meaning the record of the state of the changed event.
```
TimoGlastra, line 97: This is the structure we use in the tests. The payload of the webhook events are the serialized records. Maybe we can say something like:

> The payload of the webhook contains the serialized record related to the topic. For the `connections` topic this will be a `ConnectionRecord`, for the `credentials` topic it will be a `CredentialRecord`, and so on.

PR #93, opened 2022-02-05
feat: rest webhooks

### LINKED ISSUE #63: [rest] Add web hooks and events to rest server

Annelein:
Closes #63 
Signed-off-by: annelein <anneleinvanreijen@gmail.com>

### DIFF
--- a/packages/rest/README.md
+++ b/packages/rest/README.md
@@ -64,3 +64,34 @@ const run = async (agent: Agent) => {
 // A Swagger (OpenAPI) definition is exposed on http://localhost:3000/docs
 run()
 ```
+
+### Webhooks
+
+We use webhooks as a method for the rest api to have the option to call the controller in case of an event.
+
+Current supported events are:
+
+- `Connections`
+- `Credentials`
+- `Proofs`
+
+Example of usage:
+
+```ts
+// You can either call startServer() or setupServer() and pass the ServerConfig interface with a webhookUrl
+
+const run = async (agent: Agent) => {
+  const config = {
+    port: 3000,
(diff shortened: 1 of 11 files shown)


TimoGlastra (approved):
Nice! LGTM

PR #96, opened 2022-02-23
feat(rest): add cli and docker image publishing

TimoGlastra:
Add CLI to the rest (makes it usable without code) and also adds docker publishing. This means we don't have to set up a boilberplate project every time we use the rest server.

### DIFF
--- a/packages/rest/Dockerfile
+++ b/packages/rest/Dockerfile
@@ -1,0 +1,40 @@
+FROM ubuntu:18.04 as base
+
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN apt-get update -y && apt-get install -y \
+    software-properties-common \
+    apt-transport-https \
+    curl \
+    # Only needed to build indy-sdk
+    build-essential 
+
+# libindy
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys CE7709D068DB5E88
+RUN add-apt-repository "deb https://repo.sovrin.org/sdk/deb bionic stable"
+
+# nodejs
+RUN curl -sL https://deb.nodesource.com/setup_16.x | bash
+
+# yarn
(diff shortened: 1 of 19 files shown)


janrtvld (approved):

packages/rest/README.md
```diff
@@ -95,3 +95,34 @@ In case of an event, we will send the event to the webhookUrl with the topic of
...
+      # possible to set values using env variables
+      AFJ_REST_LOG_LEVEL: 1
+    volumes:
+      # also possible to set values using json
+      - ./myConfig.json:/config.json
+    ports:
+      - '5000:5000'
+      - '5001:5001'
+      - '3000:3000'
+    platform: linux/amd64
```
blu3beri, line 119: Will this work on m1?

packages/rest/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM ubuntu:18.04 as base
```
blu3beri, line 1: I think you wanted me to do this once and I kinda forgot, but don't we want to create an afj image with libindy and all installed?

blu3beri (approved):
I really like the addition! I do think that it could benefit from some documentation. Like a README and some minor in-code documentation.

packages/rest/README.md
TimoGlastra: Partly. It works until you try to open a connection to the ledger :(

Libindy is not available for ubuntu arm. However, BCGov is working on ARM images that include libindy I think.

packages/rest/Dockerfile
TimoGlastra: Yeah may be still worth looking at.

(here's the issue: https://github.com/animo/indy-nodejs-docker/issues/1)

TimoGlastra:
> I do think that it could benefit from some documentation. Like a README and some minor in-code documentation.

Sure will take add some documentation!

PR #111, opened 2022-04-14
fix(push-notifications): add @id to messages

TimoGlastra:
No `@id` was added to push notification messages which means AFJ would refuse to send the message due to validation errors.


TimoGlastra:
@janrtvld this fixes the `LedgerNotFound` issue you mentioned. It appareantly had something to do with the ledger id having a space in it?

PR #113, opened 2022-04-28
fix(react-hooks): update providers on delete events

Przytua:
There were no delete events in `core` package, so providers only listen to change events. Whenever a record is deleted, for example with: 
```
agent.credentials.deleteById(credentialId)
```
provider is not updated, and the `useCredentials` hook does not update returned credentials array.
This branch utilizes changes made in https://github.com/hyperledger/aries-framework-javascript/pull/716 and adds delete events listeners to fix that issue.

BREAKING CHANGE: `useConnectionByState` now needs a `DidExchangeState` state value instead of a `ConnectionState` state value.

### DIFF
--- a/packages/react-hooks/src/ConnectionProvider.tsx
+++ b/packages/react-hooks/src/ConnectionProvider.tsx
@@ -1,6 +1,6 @@
-import type { Agent, ConnectionStateChangedEvent, ConnectionRecord, DidExchangeState } from '@aries-framework/core'
+import type { Agent, ConnectionStateChangedEvent, DidExchangeState, RecordDeletedEvent } from '@aries-framework/core'
 
-import { ConnectionEventTypes } from '@aries-framework/core'
+import { ConnectionEventTypes, ConnectionRecord, RepositoryEventTypes } from '@aries-framework/core'
 import * as React from 'react'
 import { createContext, useState, useEffect, useContext, useMemo } from 'react'
 
@@ -56,7 +56,7 @@ const ConnectionProvider: React.FC<Props> = ({ agent, children }) => {
 
   useEffect(() => {
     if (!connectionState.loading) {
-      const listener = (event: ConnectionStateChangedEvent) => {
+      const stateChangedListener = (event: ConnectionStateChangedEvent) => {
         const newConnectionsState = [...connectionState.connections]
 
         const index = newConnectionsState.findIndex((connection) => connection.id === event.payload.connectionRecord.id)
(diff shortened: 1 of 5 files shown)


Przytua:
I've updated it to use generic repository events (https://github.com/hyperledger/aries-framework-javascript/issues/723)

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -27,7 +32,7 @@ export const useConnectionById = (id: string): ConnectionRecord | undefined => {
 export const useConnectionByState = (state: ConnectionState): ConnectionRecord[] => {
```
TimoGlastra, line 32: Maybe we can update this to `DidExchangeState`? That's what we're using now for all connections regardless of protocol so we can work with a singular set of states (instead of two sets with slightly different values). Then we won't need `rfc0160StateFromDidExchangeState`

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -71,10 +76,26 @@ const ConnectionProvider: React.FC<Props> = ({ agent, children }) => {
...
       }
-      agent?.events.on(ConnectionEventTypes.ConnectionStateChanged, listener)
+
+      const deletedListener = async (event: RecordDeletedEvent<ConnectionRecord>) => {
+        if (event.payload.record.type !== ConnectionRecord.type) {
+          return
+        }
+        const newConnectionsState = [
+          ...connectionState.connections.filter((connection) => connection.id != event.payload.record.id),
+        ]
```
TimoGlastra, line 86: filter already returns a new array
```suggestion
        const newConnectionsState = connectionState.connections.filter((connection) => connection.id != event.payload.record.id)
```

packages/react-hooks/src/CredentialProvider.tsx
```diff
@@ -71,10 +71,25 @@ const CredentialProvider: React.FC<Props> = ({ agent, children }) => {
         })
       }
 
-      agent?.events.on(CredentialEventTypes.CredentialStateChanged, listener)
+      const deletedListener = async (event: RecordDeletedEvent<CredentialExchangeRecord>) => {
+        if (event.payload.record.type !== CredentialExchangeRecord.type) {
+          return
+        }
+        const newCredentialsState = [
```
TimoGlastra, line 78: same here

packages/react-hooks/src/ProofProvider.tsx
```diff
@@ -68,10 +68,23 @@ const ProofProvider: React.FC<Props> = ({ agent, children }) => {
         })
       }
 
-      agent?.events.on(ProofEventTypes.ProofStateChanged, listener)
+      const deletedListener = async (event: RecordDeletedEvent<ProofRecord>) => {
+        if (event.payload.record.type !== ProofRecord.type) {
+          return
+        }
+        const newProofsState = [...proofState.proofs.filter((proof) => proof.id != event.payload.record.id)]
```
TimoGlastra, line 75: same here

TimoGlastra (commented):
Great @Przytua, can you update the version of AFJ in the package.json so we can make the CI happy? I've updated the target branch to `afj-0.2.0` so we can already get this merged

Przytua:
@TimoGlastra updated

TimoGlastra (approved):

PR #114, opened 2022-05-04
feat(react-hooks): revocation notification events

TheTreek:
Made it so Revocation Notification events will also trigger the CredentialProvider listener.

### DIFF
--- a/packages/react-hooks/package.json
+++ b/packages/react-hooks/package.json
@@ -25,13 +25,13 @@
     "test": "jest"
   },
   "devDependencies": {
-    "@aries-framework/core": "^0.1.0",
+    "@aries-framework/core": "^0.2.0-alpha.56",
     "react": "^17.0.2",
     "rimraf": "~3.0.2",
     "typescript": "~4.4.2"
   },
   "peerDependencies": {
-    "@aries-framework/core": "^0.1.0",
+    "@aries-framework/core": "^0.2.0-alpha.56",
     "react": "^17.0.2"
   }
 }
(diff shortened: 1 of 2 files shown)


TimoGlastra:
This looks good to me. However I'm going to hold it until 0.2.0 is released as otherwise we'll get in a version mess if we're going to make releases for alpha versions of AFJ (we could look into this, but would need a different release channel)

JamesKEbert:
LGTM!
@TimoGlastra, we've got a use case that needs this while using an alpha version of AFJ. Sorry, I can't recall, does merging this automatically trigger a release or is it a manual process? If it's manual, could we merge it in so that it's available for a manual TGZ creation, but wait to release until after 0.2.0 is released. 
@TheTreek, would that address our needs here you think?

TheTreek:
@TimoGlastra What would the effort to have a 2.0 branch/release look like?

TimoGlastra:
> merging this automatically trigger a release or is it a manual process?

It doesn't automatically create a new release, but it does mean we can't create bug fixes anymore for the 0.1.0 version.

> so that it's available for a manual TGZ creation,

I think you can also create a manual TGZ from a branch right?

> What would the effort to have a 2.0 branch/release look like?

It would be quite complex to create prereleases for a prelrelease of afj. The ci / releasing part would be doable, but mainly the versioning will be challenging. We would need to move this package to a 0.3.0 prerelease, and start releasing from a afj-0.2.0 branch. I don't currently have the time to work on this sadly

TimoGlastra:
I've changed the base branch to `afj-0.2.0`. If you can update this PR with latest change from main and latest alpha version of AFJ we can get this merged

TimoGlastra:
Bump @TheTreek @JamesKEbert could you update this with the latest alpha version?

PR #115, opened 2022-05-12
feat(pn)!: more services

blu3beri:
- Added apns service
- Added fcm-android service

* Will add a link to the RFCs when they are merged

BREAKING CHANGE: `PushNotificationsModule` does not exist anymore. You now have to use one that is specific to your service, e.g. `PushNotificationsApnsModule` or `PushNotificationsFcmModule`


blu3beri:
@TimoGlastra could you review this? It should be in sync again with the latest versions of the RFCs.

TimoGlastra (approved):

PR #119, opened 2022-06-14
feat(redux-store)!: update to afj 0.2.0

karimStekelenburg:
- build: bump afj core version
- feat: remove thunks and add setter actions
- feat: update CredentialRecord fields

BREAKING CHANGE: The thunks that exposed all module methods from AFJ are now removed. Instead of using the thunks, you should use the agent directly.


packages/redux-store/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@aries-framework/redux-store",
   "main": "build/index",
   "types": "build/index",
-  "version": "0.2.0",
+  "version": "0.3.0",
```
TimoGlastra, line 5: Version will be increased by the CD pipeline

packages/redux-store/src/slices/connections/connectionsSlice.ts
```diff
@@ -39,6 +37,9 @@ const connectionsSlice = createSlice({
   name: 'connections',
   initialState,
   reducers: {
+    setConnections: (state, action: PayloadAction<ConnectionRecord[]>) => {
```
TimoGlastra, line 40: Maybe we should call this `setConnectionRecords`. This way it's more consistent when we want to do `setMediations` (as it could be `setMediationRecords`)

TimoGlastra (changes requested):
Nice! Could you point this PR to the `afj-0.2.0` branch?


Is this going to be documented in the aries javascript docs? It may not be clear to all users you need to query the records yourself (while we do handle the updating)

TimoGlastra:
After thinking about it a bit more maybe we should add the getXXX thunks in (only those). We still have the listeners in place which means if a state changed event is emitted we'll update it in the store, however the store is not responsible for setting the values is the first place which seems a bit weird. What do you think? We should probably also update the listeners to use the new `RecordUpdatedEvent`, `RecordDeletedEvent` and `RecordSavedEvent` instead of the `StateChangedEvent`. Thoughts?

packages/redux-store/src/slices/connections/connectionsSlice.ts
karimStekelenburg: Fixed in f295b5c

TimoGlastra:
BTW @karimStekelenburg I've succesfully used the updated readux-store package in one of our wallets, so all seems to be working fine 👍 .

If you could make the final changes we can get this merged

PR #128, opened 2022-06-23
feat(react-hooks): revocation notification events

amanji:
Extends PR #114 

* Updates to latest AFJ version
* Fixes validation warnings
* Update credential, proof, connection and basic message provider hooks to listen for generic record events over specific state changed/revocation notification events


TimoGlastra:
I was thinking, maybe we can remove all state changed and revocation notification events and just listen for the saved/updated/deleted events and update the store based on that?

amanji:
I can update the `rxjs` operator in AFJ and further clean up the event listeners if we want.

JamesKEbert:
@amanji this is based off of @TheTreek's PR #114 ? Or is it a duplication?

amanji:
This is based off of @TheTreek PR. I checked out their commit and worked on top of it so their contributions are still included.

amanji:
I'll update this to the official release

packages/react-hooks/src/CredentialProvider.tsx
```diff
@@ -56,13 +62,29 @@ const CredentialProvider: React.FC<Props> = ({ agent, children }) => {
 
   useEffect(() => {
     if (!credentialState.loading) {
-      const stateChangedListener = async (event: CredentialStateChangedEvent) => {
+      const savedListener = async (event: RecordSavedEvent<CredentialExchangeRecord>) => {
```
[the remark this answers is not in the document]
amanji: Right! I can do that. IIRC there are the Proof and Connection Records remaining, correct? Am I missing any others?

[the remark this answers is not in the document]
TimoGlastra: I think we also have basic messages

packages/react-hooks/src/RecordProvider.tsx
```diff
@@ -0,0 +1,108 @@
...
+const RecordProvider: React.FC<Props> = ({ agent, children }) => {
+  const [recordState, setRecordState] = useState<RecordContextInterface>({
+    loading: true,
+    records: [],
+  })
+
+  const setInitialState = async () => {
+    if (agent) {
+      const records = [
+        ...((await agent.connections.getAll()) as unknown as BaseRecord[]),
```
amanji, line 43: @TimoGlastra I don't suppose there is a module for getting base records?

TimoGlastra: No we need the class and record type to correctly retrieve records from storage...

amanji: Ok. So does this approach work then? It really obviates the need for the individual providers. I have just extended the existing hooks.

amanji: That was my concern, that we can no longer subscribe to individual events or record types. Every hook now filters through every record and receives an update for any record type (I suppose this is not what we want)

amanji:
Should I reopen this against a different branch?

TimoGlastra:
Ah yes sorry, this should now be opened against main. Didn't realize your pr would be closed

PR #133, opened 2022-07-03
feat(redux-store)!: update to afj 0.2.0

TimoGlastra:
Updates redux store package to AFJ 0.2.0, removes most of the thunks (except get all to initialize the state) and updates the event listeners to the generic record events.

Supersedes #119.

MAKE SURE TO ADD THE FOLLOWING BREAKING CHANGES TO THE END OF THE SQUASHED COMMIT MESSAGE:

BREAKING CHANGE: the state changed events listeners have been updated to the new generic record saved/updated/deleted events. Instead of using a listener for each of the record types, you only have to use the `startRecordsListeners` method now.

BREAKING CHANGE: The thunks that exposed all module methods from AFJ are now removed. Instead of using the thunks, you should use the agent directly.


TimoGlastra:
> LGTM, only question I have is why you left the `createFooRecord()` methods in? convenience?

So we can manage the whole record lifecycle (retrieving and the saving/updating/deleting using listeners)

PR #136, opened 2022-07-05
refactor(react-hooks)!: update AFJ to stable 0.2.0

amanji:
This PR includes a refactor of the current react-hooks extension including:
* Updated AFJ dependency to the official 0.2.0 release
* Base record events are now handled as opposed to events for each specific record type
* Introduces a typed reducer hook factory to handle record actions in providers
* Uses agent event observable stream to allow for introduction of rxjs operators


TimoGlastra:
I'm going to leave this open for a bit to see if @JamesKEbert has any comments, but otherwise this PR is good to merge from my side

amanji:
I've gone back and forth on this a few times now and I think the implementation we have ATM is the most logical IMO. Looking at the suggested implementation for the redux extension, it benefits from being able to compose/combine reducers together, with the redux library, so it can handle actions for each record type within slices while having a single base record event listener and action dispatcher. The hooks extensions are using the simpler `useReducer` hook which would require us to implement our own "combine reducer" function or use redux functions to achieve this (which I don't think we want to do). While not ideal to have individual listeners in each provider I think we can revisit the implementation later and refactor. I'm still in favour of adding `rxjs` operators to improve stream performance and clean up the subscriptions.

packages/react-hooks/src/BasicMessageProvider.tsx
```diff
@@ -28,6 +29,7 @@ export const useBasicMessagesByConnectionId = (connectionId: string): BasicMessa
 
 interface Props {
   agent: Agent | undefined
+  children: React.ReactNode
 }
```
TimoGlastra, line 33: This is fine, but just so you know. You can do `const Component = React.FC<PropsWithChildren<Props>>` and then you don't have to declare the type for children yourself

packages/react-hooks/src/RecordProvider.tsx
```diff
@@ -0,0 +1,69 @@
+import type { BaseRecord, RecordSavedEvent, RecordDeletedEvent, RecordUpdatedEvent } from '@aries-framework/core'
+
+import { RepositoryEventTypes } from '@aries-framework/core'
+import { useReducer } from 'react'
+
+export enum RecordProviderEventTypes {
+  RecordsLoaded = 'RecordsLoaded',
+}
+
+export interface RecordsState<R extends BaseRecord<any, any, any>> {
```
TimoGlastra, line 10: There's quite some eslint warnings in this file. In the redux package I created the following type with a single eslint disable:

```
type BaseRecordAny = BaseRecord<any, any, any>
```

TimoGlastra (approved):

TimoGlastra:
It seems CI is failing

packages/react-hooks/src/BasicMessageProvider.tsx
amanji: > It seems CI is failing

Seems to be failing with the redux extension for some reason

PR #140, opened 2022-07-22
fix(redux-store): tests and proper record update

blu3beri:
- Minor cleanup for consistency
- Added some unit tests for record event changes
- Fixes an issue where the first record would always be updated


TimoGlastra (approved):
🚀

PR #146, opened 2022-08-11
feat(react-hooks): extend credential providers with format data

amanji:
Signed-off-by: Akiff Manji <akiff.manji@gmail.com>

### DIFF
--- a/packages/react-hooks/src/CredentialProvider.tsx
+++ b/packages/react-hooks/src/CredentialProvider.tsx
@@ -1,5 +1,5 @@
-import type { RecordsState } from './recordUtils'
-import type { Agent, CredentialState } from '@aries-framework/core'
+import type { CombinedRecordsState, RecordsState } from './recordUtils'
+import type { Agent, CredentialState, GetFormatDataReturn, IndyCredentialFormat } from '@aries-framework/core'
 import type { PropsWithChildren } from 'react'
 
 import { CredentialExchangeRecord } from '@aries-framework/core'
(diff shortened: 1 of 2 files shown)


packages/react-hooks/src/CredentialProvider.tsx
```diff
@@ -1,30 +1,35 @@
...
-  removeRecord,
-  updateRecord,
-  addRecord,
 } from './recordUtils'
 
-const CredentialContext = createContext<RecordsState<CredentialExchangeRecord> | undefined>(undefined)
+const CredentialContext = createContext<CombinedRecordsState<CredentialExchangeRecord> | undefined>(undefined)
 
-export const useCredentials = () => {
+export const useCombinedCredentials = () => {
```
TimoGlastra, line 20: Maybe useCredentialsWithFormatData()? Or do you think that is too long? 

combined doesn't really give any context on what is combined

TimoGlastra (approved):
Nice!

I've updated the title to be `feat` instead of `refactor` as there is a change in functionality, and otherwise a new release won't get triggered for just a refactor

packages/react-hooks/src/recordUtils.ts
```diff
@@ -14,9 +16,20 @@ import { map, filter, pipe } from 'rxjs'
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type BaseRecordAny = BaseRecord<any, any, any>
 type RecordClass<R extends BaseRecordAny> = Constructor<R> & { type: string }
+
+interface CombinedRecord<R extends BaseRecord> {
+  record: R
+  formatData: GetFormatDataReturn<[IndyCredentialFormat]>
```
blu3beri, line 22: If we want to only specifically do `IndyCredentialFormat` I think we should name it accordingly. 

I also implemented this locally for some projects and just made the formatData a separate hook.

`const formattedCredential = useCredentialFormatDataById(credentialId)`

or

`const formattedCredential = useCredentialFormatData()`

Which could then optionally take a generic for the format data, i.e. `IndyCredentialFormat`. I could also create a PR with this code (and for proofs.)

amanji:
I think this PR is probably quite out of sync with what's in AFJ today. 

IMHO if you have working code that expands on this and it's not adding unnecessary work for you I'd say maybe we go with that instead.

Let me know. I'd be happy to review once it's ready.

berendsliedrecht:
Opened #231 that supersedes this.

PR #149, opened 2022-08-23
feat(hooks): filter by multiple states with inverse

### LINKED ISSUE #147: [hooks] useCredentialsByState inverse filter
While using the useCredentialsByState hook I came across a use-case where it would be handy to have inverse filters. Instead of querying for eg `useCredentialByState('done')` it would be handy to be able to retrieve for all states states but certain state(s).
This could also be a list. 

Proposal:
* allow input param to be a list of states instead of just a single state -> `useCredentialByState(["done", "offer-sent"])`
* also allow to negate that input param list along the lines of `useCredentialByState([!"done", !"offer-sent"])` which retrieves the records for all states but 'done" and 'offer-sent'

This might also apply to proofs and connections.

morrieinmaas:
closes #147

### DIFF
--- a/packages/react-hooks/src/ConnectionProvider.tsx
+++ b/packages/react-hooks/src/ConnectionProvider.tsx
@@ -1,8 +1,8 @@
 import type { RecordsState } from './recordUtils'
-import type { Agent, DidExchangeState } from '@aries-framework/core'
+import type { Agent } from '@aries-framework/core'
 import type { PropsWithChildren } from 'react'
 
-import { ConnectionRecord } from '@aries-framework/core'
+import { DidExchangeState, ConnectionRecord } from '@aries-framework/core'
 import { useState, createContext, useContext, useEffect, useMemo } from 'react'
 import * as React from 'react'
 
@@ -30,12 +30,22 @@ export const useConnectionById = (id: string): ConnectionRecord | undefined => {
   return connections.find((c: ConnectionRecord) => c.id === id)
 }
 
-export const useConnectionByState = (state: DidExchangeState): ConnectionRecord[] => {
+export const useConnectionByState = (
+  state: DidExchangeState | DidExchangeState[],
+  invertSearch = false
+): ConnectionRecord[] => {
+  let states = typeof state === 'string' ? [state] : state
+
+  if (invertSearch) {
+    states = Object.values(DidExchangeState).filter((v) => !states.includes(v))
+  }
   const { records: connections } = useConnections()
-  const filteredConnections = useMemo(
-    () => connections.filter((c: ConnectionRecord) => c.state === state),
-    [connections, state]
-  )
+
+  const filteredConnections = states
+    .map((filterState: DidExchangeState) =>
+      useMemo(() => connections.filter((c: ConnectionRecord) => c.state === filterState), [connections, filterState])
+    )
+    .flat()
   return filteredConnections
 }
 
--- a/packages/react-hooks/src/CredentialProvider.tsx
+++ b/packages/react-hooks/src/CredentialProvider.tsx
@@ -34,17 +34,14 @@ export const useCredentialByState = (
   state: CredentialState | CredentialState[],
   invertSearch = false
 ): CredentialExchangeRecord[] => {
-  let filters: CredentialState[] = []
-  const inputStates = typeof state === 'string' ? [state] : state
+  let states = typeof state === 'string' ? [state] : state
 
   if (invertSearch) {
-    filters = Object.values(CredentialState).filter((v) => !inputStates.includes(v))
-  } else {
-    filters = inputStates
+    states = Object.values(CredentialState).filter((v) => !states.includes(v))
   }
 
   const { records: credentials } = useCredentials()
-  const filteredCredentials = filters
+  const filteredCredentials = states
     .map((filterState: CredentialState) =>
       useMemo(() => credentials.filter((c: CredentialExchangeRecord) => c.state === filterState), [credentials, state])
     )
--- a/packages/react-hooks/src/CredentialProvider.tsx
+++ b/packages/react-hooks/src/CredentialProvider.tsx
@@ -1,8 +1,8 @@
 import type { RecordsState } from './recordUtils'
-import type { Agent, CredentialState } from '@aries-framework/core'
+import type { Agent } from '@aries-framework/core'
 import type { PropsWithChildren } from 'react'
 
-import { CredentialExchangeRecord } from '@aries-framework/core'
+import { CredentialState, CredentialExchangeRecord } from '@aries-framework/core'
 import { useState, createContext, useContext, useEffect, useMemo } from 'react'
 import * as React from 'react'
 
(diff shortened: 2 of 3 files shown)


packages/react-hooks/src/CredentialProvider.tsx
```diff
@@ -30,12 +30,22 @@ export const useCredentialById = (id: string): CredentialExchangeRecord | undefi
   return credentials.find((c: CredentialExchangeRecord) => c.id === id)
 }
 
-export const useCredentialByState = (state: CredentialState): CredentialExchangeRecord[] => {
+export const useCredentialByState = (
+  state: CredentialState | CredentialState[],
+  invertSearch = false
+): CredentialExchangeRecord[] => {
+  let states = typeof state === 'string' ? [state] : state
```
blu3beri, line 37:
```suggestion
  let states = [...state]
```

or

```suggestion
  const states = inverted ? Object.values(CredentialState).filter((v) => ![...state].includes(v)) : [...state]
```

But that might become a bit unreadable.

blu3beri (changes requested):
Nice addition! Could you also add this for the other `useXByState` functions?

packages/react-hooks/src/CredentialProvider.tsx
morrieinmaas: Thanks for the feedback! the top part is good imho

morrieinmaas:
> Nice addition! Could you also add this for the other `useXByState` functions?

Yes good shout. was thinking that. should I do that in this PR or open another one?

blu3beri:
> > Nice addition! Could you also add this for the other `useXByState` functions?
> 
> 
> 
> Yes good shout. was thinking that. should I do that in this PR or open another one?

This PR would be fine.

packages/react-hooks/src/CredentialProvider.tsx
morrieinmaas: ah just a side note: `let states = [...state]` won't work. when state is a string it will create a character array of each char of the string instead an array with the one entry being the string.

morrieinmaas: so not sure I can think of another/better way to get rid of the let then

morrieinmaas: happy for suggestions

blu3beri: Ah I was not aware spread operators work like that on strings, makes sense. 

Just leave it as-is then.

morrieinmaas: I've added it to the other byId stuff. Lemme know if you have anymore feedback

blu3beri (approved):
LGTM! Just as a minor note: stuff like `invertSearch` really needs external documentation. Maybe have to bring it up again during the WG call.

morrieinmaas:
> LGTM! Just as a minor note: stuff like `invertSearch` really needs external documentation. Maybe have to bring it up again during the WG call.

That's an excellent shout! - and agreed. I'll bring it up Thursday. On the plus side this only extends and doesn't change the previous behaviour.

genaris:
> LGTM! Just as a minor note: stuff like `invertSearch` really needs external documentation. Maybe have to bring it up again during the WG call.

Agree! This addition is pretty useful. And it's true that the `invertSearch` might not be pretty intuitive at a first sight, but if you are looking for "records not in state" you'll certainly understand its purpose.

However, using a separate method for extenal interface could be clearer as well, although I don't know how to name it in a fancy way. `useXNotInState` ? `useXInStateDifferentFrom` ?

blu3beri:
> > LGTM! Just as a minor note: stuff like `invertSearch` really needs external documentation. Maybe have to bring it up again during the WG call.
> 
> Agree! This addition is pretty useful. And it's true that the `invertSearch` might not be pretty intuitive at a first sight, but if you are looking for "records not in state" you'll certainly understand its purpose.
> 
> Defining a separate method for this purpose in the external interface could be clearer as well, although I don't know how to name it in a fancy way. `useXNotInState` ? `useXInStateDifferentFrom` ?

I am okay with keeping the API like this, but it was more just a general point that I would really like to "enforce" proper TSDoc on public functions. It can help so much and we as maintainers just have to pay a bit more attention that they are kept up-to-date.

morrieinmaas:
> > > LGTM! Just as a minor note: stuff like `invertSearch` really needs external documentation. Maybe have to bring it up again during the WG call.
> > 
> > 
> > Agree! This addition is pretty useful. And it's true that the `invertSearch` might not be pretty intuitive at a first sight, but if you are looking for "records not in state" you'll certainly understand its purpose.
> > Defining a separate method for this purpose in the external interface could be clearer as well, although I don't know how to name it in a fancy way. `useXNotInState` ? `useXInStateDifferentFrom` ?
> 
> I am okay with keeping the API like this, but it was more just a general point that I would really like to "enforce" proper TSDoc on public functions. It can help so much and we as maintainers just have to pay a bit more attention that they are kept up-to-date.

for sure. agreed

TimoGlastra:
I quite like the `useXNotInState` suggestion, as it removes the need for a invert boolean. One of the lessons I've learned from the clean code book is that switch booleans can almost always be prevented by using two methods that do the inverse. 

Especially since we don't use an object to pass the param, the boolean is quite opaque in it's working. If you see code like this, my first guess (without looking at docs, etc..) would still be that all credentials with state done would be returned:

```ts
useCredentialsByState([CredentialState.Done], true)
```

While with a different method it would be clear that that's not the case:

```ts
useCredentialsNotInState([CredentialState.Done])
```

morrieinmaas:
> I quite like the `useXNotInState` suggestion, as it removes the need for a invert boolean. One of the lessons I've learned from the clean code book is that switch booleans can almost always be prevented by using two methods that do the inverse.
> 
> Especially since we don't use an object to pass the param, the boolean is quite opaque in it's working. If you see code like this, my first guess (without looking at docs, etc..) would still be that all credentials with state done would be returned:
> 
> ```ts
> useCredentialsByState([CredentialState.Done], true)
> ```
> 
> While with a different method it would be clear that that's not the case:
> 
> ```ts
> useCredentialsNotInState([CredentialState.Done])
> ```

@TimoGlastra nice idea. I'll have a go at that. thanks

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -30,12 +30,22 @@ export const useConnectionById = (id: string): ConnectionRecord | undefined => {
...
-  const filteredConnections = useMemo(
-    () => connections.filter((c: ConnectionRecord) => c.state === state),
-    [connections, state]
-  )
+
+  const filteredConnections = states
+    .map((filterState: DidExchangeState) =>
+      useMemo(() => connections.filter((c: ConnectionRecord) => c.state === filterState), [connections, filterState])
+    )
+    .flat()
```
TimoGlastra, line 48: Also this loops through the connections list for each of the passed states. What if we filter it like this: 

```
const filteredConnections = useMemo(() => connections.filter(c => states.includes(c.state)))
```

This can save quite some loops if we're filtering for multiple states (especially with the inverse filter)

packages/react-hooks/src/CredentialProvider.tsx
```diff
@@ -30,12 +30,25 @@ export const useCredentialById = (id: string): CredentialExchangeRecord | undefi
...
+  if (invertSearch) {
+    states = Object.values(CredentialState).filter((v) => !states.includes(v))
+  }
+
   const { records: credentials } = useCredentials()
-  const filteredCredentials = useMemo(
-    () => credentials.filter((c: CredentialExchangeRecord) => c.state === state),
-    [credentials, state]
-  )
+  const filteredCredentials = states
```
TimoGlastra, line 44: Same here, I think we can optimize this

packages/react-hooks/src/ProofProvider.tsx
```diff
@@ -30,9 +30,19 @@ export const useProofById = (id: string): ProofRecord | undefined => {
...
+export const useProofByState = (state: ProofState | ProofState[], invertSearch = false): ProofRecord[] => {
+  let states = typeof state === 'string' ? [state] : state
+
+  if (invertSearch) {
+    states = Object.values(ProofState).filter((v) => !states.includes(v))
+  }
+
   const { records: proofs } = useProofs()
-  const filteredProofs = useMemo(() => proofs.filter((p: ProofRecord) => p.state === state), [proofs, state])
+  const filteredProofs = states
```
TimoGlastra, line 41: Same

packages/react-hooks/src/ConnectionProvider.tsx
blu3beri: Don't forget here to add the dependency array :).

PR #151, opened 2022-08-25
fix(react-hooks): update react peer dependency

JamesKEbert:
This updates the React peer dependency to 18.0.0 as part of an effort to update Aries Bifold to newer versions of React Native. 

This would be considered a breaking change according to Semver due to the peer dependency change (via https://github.com/semver/semver/issues/502)

### DIFF
--- a/packages/react-hooks/package.json
+++ b/packages/react-hooks/package.json
@@ -35,6 +35,6 @@
   },
   "peerDependencies": {
     "@aries-framework/core": "^0.2.0",
-    "react": "^18.0.0"
+    "react": ">=17.0.0 <19.0.0"
   }
 }
--- a/packages/react-hooks/package.json
+++ b/packages/react-hooks/package.json
@@ -29,12 +29,12 @@
   },
   "devDependencies": {
     "@aries-framework/core": "^0.2.0",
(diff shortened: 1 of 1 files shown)


packages/react-hooks/package.json
```diff
@@ -29,12 +29,12 @@
   },
   "devDependencies": {
     "@aries-framework/core": "^0.2.0",
-    "react": "^17.0.2",
+    "react": "^18.0.0",
```
TimoGlastra, line 32: What if we make it a semver range 

E.g.:

```
>=17.0.0 <=19.0.0
```

or just

```
>=17.0.0
```

Then it doesn't have to be a breaking change

TimoGlastra (changes requested):

packages/react-hooks/package.json
JamesKEbert: I don't see any reason that would not work! You are a wise man Timo, this hadn't occurred to me 👏 I'll go ahead and update this shortly.

TimoGlastra (approved):

JamesKEbert:
Ahh thanks for catching the 'chore' vs 'fix' issue :)

PR #152, opened 2022-08-29
refactor(react-hooks): filter method

morrieinmaas:


packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -30,22 +30,29 @@ export const useConnectionById = (id: string): ConnectionRecord | undefined => {
...
+export const useConnectionByState = (state: DidExchangeState | DidExchangeState[]): ConnectionRecord[] => {
+  const states = typeof state === 'string' ? [state] : state
+
+  const { records: connections } = useConnections()
+
+  const filteredConnections = connections.filter((r: ConnectionRecord) =>
+    useMemo(() => {
+      if (states.includes(r.state)) return r
+    }, [connections])
+  )
```
TimoGlastra, line 42: I think we should switch the memo and filter so the memo is on the outside. This way we will remember the whole filteredConnections filter as long as the connections or the state don't change. Also we should add the states array to the dependency array.

```suggestion
  const filteredConnections = useMemo(() => connections.filter((connection) => states.includes(connection)), [connections, states])
```


In addition there are a few things to take into account with useMemo. useMemo will recalculate the value if any of the values in the dependency array change. If we only pass one state to the `useConnectionByState` method, on every call it will be turned into an array (first line of the method). As the reference of the array is different on every call the useMemo will never use the cached value and re-calculate the value each time.

In addition if we call `useConnectionByState` with an array of states inline (see example below) the array will be different on every call and we still never will use the cached version of the useMemo method.

```ts
const connections = useConnectionByState([ConnectionState.Done, ConnectionState.Response])
```


So I see a few things we can do to mitigate these issues. They can be used together:
1. Also use useMemo for the states property. This way we'll only create the value once the input has changed:
```ts
const states = useMemo(() => typeof state === 'string' ? [state] : state, [state])
```
2. Call `useConnectionByState` with a stable reference. This is not something we can enforce, but somethign we should document and see as best practice. Example:
```ts
const states = [ConnectionState.Done, ConnectionState.Response]

const MyComponent = () => {
  // bad: state array changes on every call, results won't be memoized
  const myBadConnections = useConnectionByState([ConnectionState.Done, ConnectionState.Response])

  // good: state array doesn't change as it is defined outside of the render. The states value will always be the same.
  const myGoodConnections = useConnectionByState(states)

  return <Text>You have {myGoodConnections.length} connections</Text>
}
```

TimoGlastra: Also, was this method already called useConnectionInState (singular)? We should probably rename it to useConnectionsInState (plural). However not for now as that will be a breaking change

morrieinmaas: aight thx for all the feedback. I'll have a go at it tonight

TimoGlastra (approved):

PR #166, opened 2022-09-09
feat(rest): added filters to getAllCredentials

rapaktech:


packages/rest/src/controllers/credentials/CredentialController.ts
```diff
@@ -31,8 +31,17 @@ export class CredentialController extends Controller {
    */
   @Example<CredentialExchangeRecordProps[]>([CredentialExchangeRecordExample])
   @Get('/')
-  public async getAllCredentials() {
-    const credentials = await this.agent.credentials.getAll()
+  public async getAllCredentials(
+    @Query('threadId') threadId?: string,
+    @Query('connectionId') connectionId?: string,
+    @Query('state') state?: string,
```
TimoGlastra, line 37: Does it work if we use `CredentialState` as the type here? Maybe that'll give better validation and documentation in the swagger ui?

TimoGlastra: image

packages/rest/src/controllers/credentials/CredentialController.ts
```diff
@@ -31,8 +31,17 @@ export class CredentialController extends Controller {
...
+  ) {
+    let credentials = await this.agent.credentials.getAll()
+
+    if (state) credentials = credentials.filter((c) => c.state === state)
+    if (threadId) credentials = credentials.filter((c) => c.threadId === threadId)
```
TimoGlastra, line 42: Ah I see this is also done like this in the connections controller, but we should actually filter in the query, to not retrieve records that aren't needed.

We should expose this in AFJ (I'll create an issue for this), but for now we can do it like this:

```ts
    const credentialRepository = this.agent.dependencyManager.resolve(CredentialRepository)

    const credentials = await credentialRepository.findByQuery({
      connectionId,
      threadId,
      state,
    })

    return credentials.map((c) => c.toJSON())
```

When this is merged, could you also make a PR to update this for the connections controller?

@janrtvld FYI

packages/rest/tests/credential.test.ts
```diff
@@ -39,6 +39,48 @@ describe('CredentialController', () => {
     })
   })
 
+  describe('Get all credentials by state', () => {
+    test('should return all credentials', async () => {
```
TimoGlastra, line 43: As we mock the results, we aren't testing whether te query is correctly applied here.

I think we should spy on the credential repository and see if it was called with the correct parameters:

```ts
  describe('Get all credentials by state', () => {
    test('should return all credentials', async () => {
      const credentialRepository = bobAgent.dependencyManager.resolve(CredentialRepository)
      const findByQuerySpy = jest.spyOn(credentialRepository, 'findByQuery').mockResolvedValueOnce([testCredential])

      const response = await request(app).get('/credentials').query({ state: testCredential.state })

      expect(findByQuerySpy).toBeCalledWith({
        state: testCredential.state,
      })

      expect(response.statusCode).toBe(200)
      expect(response.body).toEqual([testCredential].map(objectToJson))
    })
  })
```

TimoGlastra (changes requested):
Awesome. Tested locally and it works good! Left some notes on how to make it a bit more efficient when working with large amounts of records

packages/rest/src/controllers/credentials/CredentialController.ts
TimoGlastra: https://github.com/hyperledger/aries-framework-javascript/issues/1022

packages/rest/tests/credential.test.ts
rapaktech: Got it! I'll make the changes now

TimoGlastra (approved):
🚀 🚀 🚀

TimoGlastra (approved):
🚀

TimoGlastra:
DCO is failing for some commits (seems the last two). 

Can you sign off these commits by following these instructions: https://github.com/hyperledger/aries-rfcs/blob/main/contributing.md#how-to-sign-off-previous-commits? Then I'll merge

PR #169, opened 2022-09-15
feat(rest): added create-offer endpoint and tests

rapaktech:


packages/rest/src/controllers/credentials/CredentialController.ts
```diff
@@ -157,6 +158,31 @@ export class CredentialController extends Controller {
...
+   */
+  @Example<CredentialExchangeRecordProps>(CredentialExchangeRecordExample)
+  @Post('/create-offer')
+  public async createOffer(
+    @Body() options: CreateOfferOptions,
+    @Res() internalServerError: TsoaResponse<500, { message: string }>
+  ) {
+    try {
+      const offer = await this.agent.credentials.createOffer(options)
+      await this.agent.oob.createInvitation({ messages: [offer?.message] })
```
TimoGlastra, line 176: Ah wait, this shouldn't call create invitation. There's multiple ways to create the invitation, and also a legacy way. So what we've done is first call create offer endpoint, and you the. Manually need to call create (legacy-connection less) invitation endpoint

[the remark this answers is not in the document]
rapaktech: I wasn't sure if it was

rapaktech: Okay

rapaktech: When you say endpoint, you mean the method, correct?

TimoGlastra (approved):

PR #170, opened 2022-09-19
feat(rest): added WebSocket event server

rapaktech:

### DIFF
--- a/packages/rest/package.json
+++ b/packages/rest/package.json
@@ -42,7 +42,7 @@
     "tslog": "^3.3.3",
     "tsoa": "^4.1.2",
     "tsyringe": "^4.7.0",
-    "ws": "^8.8.1",
+    "ws": "^7.5.3",
     "yargs": "^17.3.1"
   },
   "devDependencies": {
--- a/packages/rest/package.json
+++ b/packages/rest/package.json
@@ -42,6 +42,7 @@
     "tslog": "^3.3.3",
     "tsoa": "^4.1.2",
     "tsyringe": "^4.7.0",
+    "ws": "^8.8.1",
     "yargs": "^17.3.1"
   },
   "devDependencies": {
@@ -60,5 +61,9 @@
     "ts-jest": "^27.0.7",
     "ts-node-dev": "^2.0.0",
     "typescript": "^4.7.3"
+  },
+  "optionalDependencies": {
+    "bufferutil": "^4.0.6",
+    "utf-8-validate": "^5.0.9"
   }
 }
--- a/packages/rest/package.json
+++ b/packages/rest/package.json
@@ -61,9 +61,5 @@
     "ts-jest": "^27.0.7",
     "ts-node-dev": "^2.0.0",
     "typescript": "^4.7.3"
-  },
-  "optionalDependencies": {
-    "bufferutil": "^4.0.6",
-    "utf-8-validate": "^5.0.9"
   }
 }
--- a/packages/rest/package.json
+++ b/packages/rest/package.json
@@ -32,6 +32,7 @@
   "dependencies": {
     "@aries-framework/core": "^0.2.3",
     "@aries-framework/node": "^0.2.3",
+    "@types/ws": "^7.4.7",
     "body-parser": "^1.20.0",
     "cors": "^2.8.5",
     "express": "^4.18.1",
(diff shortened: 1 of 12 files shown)


packages/rest/package.json
```diff
@@ -32,6 +32,7 @@
   "dependencies": {
     "@aries-framework/core": "^0.2.3",
     "@aries-framework/node": "^0.2.3",
+    "@types/ws": "^7.4.7",
```
TimoGlastra, line 35: Don't we also need to add the `ws` package?

packages/rest/src/events/WebSocketEvents.ts
```diff
@@ -0,0 +1,95 @@
...
+) => {
+  stream.on('message', function message(data) {
+    server.clients.forEach((client) => {
+      if (client.readyState === WebSocket.OPEN) {
+        client.send(data)
+      }
+    })
+  })
+
+  agent.events.on(
```
TimoGlastra, line 34: We already have a listener for the webhook events, I think we can reuse those listener also for the websocket events (in the other events files).

Also, would this mean a webhook is sent twice to the webhookUrl now?

packages/rest/src/events/WebSocketEvents.ts
```diff
@@ -0,0 +1,95 @@
...
+        await sendWebhookEvent(config.webhookUrl + '/basic-messages', body, agent.config.logger)
+      }
+
+      server.clients.forEach((client) => {
+        if (client.readyState === WebSocket.OPEN) {
+          client.send(body)
+        } else if (client.readyState === WebSocket.CLOSED) {
+          client.terminate()
+        }
+      })
```
TimoGlastra, line 50: Maybe we can generalize this code a bit? We have the sendWebhookEvent method already, what if we create a single method called `emitEventToClient` that will a) send the webhook event to the specified url and b) send a websocket event over the existing websocket clients.

This way the type of the event, and emitting the event to the client is a bit more separated

packages/rest/src/server.ts
```diff
@@ -31,6 +36,30 @@ export const setupServer = async (agent: Agent, config: ServerConfig) => {
...
+  agent.registerOutboundTransport(httpOutboundTransport)
+  agent.registerInboundTransport(wsInboundTransport)
+  agent.registerOutboundTransport(wsOutboundTransport)
+
+  await agent.initialize()
+
+  httpInboundTransport.server?.on('upgrade', (request, socket, head) => {
+    socketServer.handleUpgrade(request, socket as Socket, head, (socket) => {
+      socketServer.emit('connection', socket, request)
+    })
```
TimoGlastra, line 60: The rest package adds a rest api around the agent, but doesn't configure the agent for you. This allows to use your own agent with the rest package.

I think we should reuse the server that we are using for the rest api (so the `server` variable) for the weboscket server.

```
  socketServer.on('connection', (stream) => {
    webSocketEvents(socketServer, stream, agent, config)
  })

  server.on('upgrade', (request, socket, head) => {
    socketServer.handleUpgrade(request, socket as Socket, head, (socket) => {
      socketServer.emit('connection', socket, request)
    })
```

TimoGlastra (changes requested):
This is a great start @rapaktech! 

Could you also add some tests that tests whether the websocket events are emitted?

packages/rest/package.json
```diff
@@ -41,6 +42,7 @@
     "tslog": "^3.3.3",
     "tsoa": "^4.1.2",
     "tsyringe": "^4.7.0",
+    "ws": "^8.8.1",
```
TimoGlastra, line 45:
```suggestion
    "ws": "^7.5.3",
```

packages/rest/package.json
```diff
@@ -59,5 +61,9 @@
     "ts-jest": "^27.0.7",
     "ts-node-dev": "^2.0.0",
     "typescript": "^4.7.3"
+  },
+  "optionalDependencies": {
+    "bufferutil": "^4.0.6",
+    "utf-8-validate": "^5.0.9"
```
TimoGlastra, line 67:
```suggestion
```

packages/rest/src/index.ts
```diff
@@ -2,11 +2,26 @@ import 'reflect-metadata'
 
 import type { ServerConfig } from './utils/ServerConfig'
 import type { Agent } from '@aries-framework/core'
+import type { Socket } from 'net'
 
+import { Server } from 'ws'
+
+// import { emitEventToClient } from './events/WebSocketEvents'
```
TimoGlastra, line 9:
```suggestion
```

TimoGlastra (approved):

PR #172, opened 2022-09-30
feat(rest): added did resolver endpoint and tests

rapaktech:


packages/rest/src/controllers/did/DidController.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  /**
+   * Retrieve connection record by connection id
+   * @param did Connection identifier
+   * @returns DidResolutionMetadata
+   */
+  @Example<DidResolutionResultProps>(DidRecordExample)
+  @Get('/:did')
+  public async getConnectionById(@Path('did') did: Did) {
```
TimoGlastra, line 28: The method name and documentation is still from the connection controller

packages/rest/src/controllers/did/DidController.ts
```diff
@@ -0,0 +1,38 @@
...
+    const didRecord = await this.agent.dids.resolve(did)
+
+    if (!didRecord.didDocument) {
```
TimoGlastra, line 31: What are the different statusses? MYbe we can differ between 404 (not found) and other error (500)

packages/rest/src/controllers/did/DidController.ts
```diff
@@ -0,0 +1,38 @@
...
+      this.setStatus(500)
+      return { didRecord }
+    }
+
+    return { ...didRecord, didDocument: objectToJson(didRecord.didDocument) }
```
TimoGlastra, line 36: I think we can just call didDocument.toJson()

packages/rest/tests/did.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+describe('DidController', () => {
+  let app: Server
+  let aliceAgent: Agent
+  let didRecord: DidResolutionResult
+
+  beforeAll(async () => {
+    aliceAgent = await getTestAgent('Connection REST Agent Test Alice', 3999)
+    app = await startServer(aliceAgent, { port: 3000 })
+    didRecord = await objectToJson(getTestDidRecord())
```
TimoGlastra, line 18: We don't need to create a did record here. The did resolver dynamically resolves the document, and isn't actually using the storage. We can just use the did document itself

packages/rest/tests/did.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  })
+
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  describe('Get did resolution result by did', () => {
+    test('should return did resolution record', async () => {
+      const spy = jest.spyOn(aliceAgent.dids, 'resolve').mockResolvedValueOnce(didRecord)
```
TimoGlastra, line 27: We don't have to mock this. The did key value can be resolved without needing a mock

TimoGlastra (commented):
Nice! Some small suggestions

packages/rest/src/controllers/did/DidController.ts
rapaktech: Right! Forgot to change it. I'll modify them now

packages/rest/src/controllers/did/DidController.ts
rapaktech: The 500 is when didDocument is empty. I'm following your recommendation from the issue description:

rapaktech: The 500 is when didDocument is empty. I'm following your recommendation from the issue description:

packages/rest/src/controllers/did/DidController.ts
rapaktech: didDocument.toJson() does not exist on the document, due to the new interface I had to extend

packages/rest/tests/did.test.ts
rapaktech: Got it

packages/rest/tests/did.test.ts
rapaktech: Got it

packages/rest/src/controllers/did/DidController.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+   * Retrieve connection record by connection id
+   * @param did Connection identifier
+   * @returns DidResolutionMetadata
+   */
+  @Example<DidResolutionResultProps>(DidRecordExample)
+  @Get('/:did')
+  public async getConnectionById(@Path('did') did: Did) {
+    const didRecord = await this.agent.dids.resolve(did)
```
TimoGlastra, line 28: The resolve method does not return a record

```suggestion
    const resolveResult = await this.agent.dids.resolve(did)
```

packages/rest/src/controllers/did/DidController.ts
TimoGlastra: The didDocument on the `.resolve()` method does have the .toJSON() method:

```
result.didDocument.toJSON()
```

packages/rest/tests/did.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    jest.clearAllMocks()
+  })
+
+  describe('Get did resolution result by did', () => {
+    test('should give 200 when did resolution record is found', async () => {
+      const did = 'did:key:z6Mkk7yqnGF3YwTrLpqrW6PGsKci7dNqh1CjnvMbzrMerSeL'
+      const response = await request(app).get(`/dids/${did}`)
+
+      expect(response.statusCode).toBe(200)
+      expect(response.body.didDocument.id).toBe(did)
```
TimoGlastra, line 29: I think it would be good to check the rest of the did document here, so we know for certain the json serialization correctly serializes the model properties to the other properties (e.g. `context` to `@context`). You can just copy the response once and test if it equals the result

rapaktech: Got it

rapaktech:
What about this @TimoGlastra @janrtvld?

jest.config.base.ts
```diff
@@ -1,7 +1,7 @@
 import type { Config } from '@jest/types'
 
 const config: Config.InitialOptions = {
-  testTimeout: 120000,
+  testTimeout: 12000,
```
TimoGlastra, line 4:
```suggestion
  testTimeout: 120000,
```

packages/rest/jest.config.ts
```diff
@@ -6,7 +6,7 @@ const config: Config.InitialOptions = {
   ...base,
   name: '@aries-framework/rest',
   displayName: '@aries-framework/rest',
-  testTimeout: 120000,
+  testTimeout: 12000,
```
TimoGlastra, line 9: My bad, these should have been changed back

```suggestion
  testTimeout: 120000,
```

packages/rest/tests/did.test.ts
```diff
@@ -0,0 +1,79 @@
...
+          'did:key:z6Mkk7yqnGF3YwTrLpqrW6PGsKci7dNqh1CjnvMbzrMerSeL#z6Mkk7yqnGF3YwTrLpqrW6PGsKci7dNqh1CjnvMbzrMerSeL',
+        ],
+        capabilityDelegation: [
+          'did:key:z6Mkk7yqnGF3YwTrLpqrW6PGsKci7dNqh1CjnvMbzrMerSeL#z6Mkk7yqnGF3YwTrLpqrW6PGsKci7dNqh1CjnvMbzrMerSeL',
+        ],
+      })
+    })
+
+    test('should give 500 when did document record is not found', async () => {
+      // const response = await request(app).get(`/dids/did:key:z6Mkk7yqnGF3YwTrLpqrW6PGsKci7dNqh1CjnvMbzrMerSeL`)
```
TimoGlastra, line 68:
```suggestion
```

TimoGlastra (approved):
Some small suggestions that can be fixed from the GH ui, but otherwise good to merge!

rapaktech:
Last merge from main seems to have broken some things

rapaktech:
Screenshot from 2022-10-07 13-01-20

From the last CI run

PR #173, opened 2022-09-30
feat(react-hooks): filter mediator connections

ryankoch13:
React Hook updates to use new connectionType property and filter by / with them. Contributed by @KolbyRKunz and @TheTreek


packages/react-hooks/package.json
```diff
@@ -28,13 +28,13 @@
     "rxjs": "^7.2.0"
   },
   "devDependencies": {
-    "@aries-framework/core": "^0.2.0",
+    "@aries-framework/core": "^0.2.5-alpha.3",
```
TimoGlastra, line 31: We keep extensions aligned with the stable releases. If you want this feature available in an extension we can make a new release

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,14 +15,71 @@ import {
...
+          }
+          for (const type of recordTypes) {
+            if (options.excludedTypes?.includes(type)) return false
+          }
+          return true
+        }),
+      [connections, options.excludedTypes]
+    )
+    connections = filteredByType
+  }
```
TimoGlastra, line 65: I think we can simplify this by creating two filters that filter based on their own criteria. This way the excludedTypes filter doesn't need to know about the state filter and vice versa.

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,14 +15,71 @@ import {
...
+
+  return { ...connectionContext, records: connections }
+}
+
+export const useConnectionsByType = (type: [ConnectionType | string]) => {
+  const { records: connections } = useConnections()
+  const filteredConnections = useMemo(
+    () =>
+      connections.filter((record: ConnectionRecord) => {
+        const recordTypes = record.getTag('connectionType') as [string]
```
TimoGlastra, line 75: `[string]` means an array with one string item in it, while `string[]` means an array of strings. I think you mean the latter here?

```suggestion
        const recordTypes = record.getTag('connectionType') as string[]
```

TimoGlastra (commented):
Can you remove the package-lock file?

Also -- If we update the required minimum version of AFJ to 0.2.5 instead of 0.2.0 that is a breaking change. Fine to make a breaking change, but just so we know that the minimum required AFJ version from now on will be 0.2.5

packages/react-hooks/package.json
TimoGlastra: Let me know if you want me to make a new release

ryankoch13: That would be amazing if we could make a new release with the connection types and use that, let us know if there's anything we can do to help with that

packages/react-hooks/src/ConnectionProvider.tsx
ryankoch13: We were trying to give users the ability to filter by both connection type and connection state at the same time with a single hook as we didn't see a clean way to do that if they are separate hooks. It should only filter by both if the options indicate to do so, and will default to doing neither.

TimoGlastra, line 65: If we write it like this, it's only a single filter, which scales a bit better if new filter type are added (just add another if statement):

```suggestion
  // Only run this code if we don't have options.excludedTypes or options.connectionState to limit amount of loops
  if (!options.connectionState && !options.excludedTypes) return connections
  
  // Filter connections based on provided filters
  return useMemo(() => connections.filter((connection) => {
   // By default we include this connection
   let valid = true
   
   
   // Filter by state (if connectionState is defined)
    if (options.connectionState) valid = c.state === options.connectionState
    
    // Exclude records with certain connection types (if defined)
    const connectionTypes = record.getTag('connectionType') as string[] | null
    if (options.excludedTypes && recordTypes) {
      valid = connectionTypes.some(connectionType => options.excludedTypes.includes(connectionType))
    }
    
    return valid
  }), [connections, options.connectionState, options.excludedTypes])
```

TimoGlastra: Also I'm not sure whether this will work, as we're using a conditional hook here, which is not allowed according to the rules of hooks. https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

So we should probably model it like this:

```ts  
  return useMemo(() => {
      // do not filter if not filter options are provided to save on a loop
      if (!options.connectionState && !options.excludedTypes) return connections

       return connections.filter((connection) => {
	   // By default we include this connection
	   let valid = true
	   
	   
	   // Filter by state (if connectionState is defined)
	    if (options.connectionState) valid = c.state === options.connectionState
	    
	    // Exclude records with certain connection types (if defined)
	    const connectionTypes = record.getTag('connectionType') as string[] | null
	    if (options.excludedTypes && recordTypes) {
	      valid = connectionTypes.some(connectionType => options.excludedTypes.includes(connectionType))
	    }
	    
	    return valid
	  })
     }, [connections, options.connectionState, options.excludedTypes])
```

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,14 +15,71 @@ import {
...
+export type useConnectionsOptions = {
+  excludedTypes?: [ConnectionType | string]
+  connectionState?: DidExchangeState
+}
+
 const ConnectionContext = createContext<RecordsState<ConnectionRecord> | undefined>(undefined)
 
-export const useConnections = () => {
+/**
+ * This method retreives the connection context for the current agent.
```
TimoGlastra, line 26:
```suggestion
 * This method retrieves the connection context for the current agent.
```

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,14 +15,71 @@ import {
...
+      [connections, options.connectionState]
+    )
+    connections = filteredByState
+  }
+
+  if (options.excludedTypes) {
+    const filteredByType = useMemo(
+      () =>
+        connections.filter((record: ConnectionRecord) => {
+          const recordTypes = (record.getTag('connectionType') as [string] | null) || []
```
amanji, line 52: Is it easier to just do:

```js
const type = (r.getTag('connectionType') || []) as [string]
```

Since you are short circuiting to an empty array anyway?

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,14 +15,71 @@ import {
...
+          // Filter out by state if needed
+          if (options.connectionState && options.connectionState !== record.state) {
```
amanji, line 54: Can this not be simplified to:

```js
 if (options.connectionState !== record.state) {
```

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,49 +15,58 @@ import {
...
+    // do not filter if not filter options are provided to save on a loop
+    if (!options.connectionState && !options.excludedTypes) return connections
+
+      return connections.filter((record: ConnectionRecord) => {
+    // By default we include this connection
+    let valid = true
+    
+    
+    // Filter by state (if connectionState is defined)
+    if (options.connectionState) valid = record.state === options.connectionState
```
JamesKEbert, line 52: Should this not just immediately return false if it doesn't match? If it doesn't match the state and it sets it to false _and_ then record types aren't excluded, valid will then get set to true, right?

packages/react-hooks/src/ConnectionProvider.tsx
JamesKEbert: Bump on this last comment @ryankoch13 ^

packages/react-hooks/src/ConnectionProvider.tsx
TimoGlastra: You're right, good catch 😄

Another option would be to do `valid = valid && <check>`

JamesKEbert:
> Also, it seems this removed useConnectionByState which means it'll be a breaking change. Is that intended?

It was, however, we ended up discussing this on the Bifold call this morning--we're going to add backwards-compatibility with the existing hooks and mark them as deprecated (as this will provide the smoothest update/upgrade strategy for devs.

packages/react-hooks/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@aries-framework/react-hooks",
   "main": "build/index",
   "types": "build/index",
-  "version": "0.4.0",
+  "version": "0.4.1",
```
TimoGlastra, line 5: This will be done automatically by the release pipeline

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,21 +15,57 @@ import {
   addRecord,
 } from './recordUtils'
 
+export type useConnectionsOptions = {
```
TimoGlastra, line 18: Types are generally PascalCase

```suggestion
export type UseConnectionsOptions = {
```

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,21 +15,57 @@ import {
...
+  excludedTypes?: [ConnectionType | string]
```
TimoGlastra, line 19: `[ConnectionType | string]` means an array of length one with ConnectionType/string. While `Array<ConnectionType | string>` means an array with ConnectionType/string

```suggestion
  excludedTypes?: Array<ConnectionType | string>
```

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,21 +15,57 @@ import {
...
+      throw new Error('useConnections must be used within a ConnectionContextProvider')
+    }
+    // do not filter if not filter options are provided to save on a loop
+    if (!options.connectionState && !options.excludedTypes) return connections
+
+    return connections.filter((record: ConnectionRecord) => {
+      // By default we include this connection
+      
+      // Filter by state (if connectionState is defined)
+      if (options.connectionState !== record.state) return false
```
TimoGlastra, line 47: This will always return false if you don't provide a connectionState to the options, while it should return true I think

```suggestion
      // Filter by state (if connectionState is defined)
      if (options.connectionState && options.connectionState !== record.state) return false
```

TimoGlastra: Might be good to add some tests? As this seems like a tricky hook

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,21 +15,57 @@ import {
...
+      
+      // Exclude records with certain connection types (if defined)
+      const recordTypes = record.connectionTypes as ConnectionType[] | string[] | null
```
TimoGlastra, line 50: It could contains string and ConnectionType items at the same time:

```suggestion
      const recordTypes = record.connectionTypes as Array<ConnectionType | string> | null
```

packages/react-hooks/src/ConnectionProvider.tsx
```diff
@@ -15,21 +15,57 @@ import {
...
+      if (options.excludedTypes && recordTypes) {
+        return recordTypes.some(connectionType => !options.excludedTypes?.includes(connectionType))
```
TimoGlastra, line 52:
```suggestion
        return recordTypes.some(connectionType => !options.excludedTypes.includes(connectionType))
```

ryankoch13: This looks like it should work but for some reason TypeScript is giving me warnings here, even if I explicitly check if excludedTypes is undefined and return right above it. Not sure what's going on here, but I changed the question to an exclamation at least.

packages/react-hooks/src/ConnectionProvider.tsx
ryankoch13: Agreed, though would it be okay if we added tests in a separate PR? I did at least run bifold with the current iteration after adding this check and it appears to be working as intended.

TimoGlastra (approved):

PR #174, opened 2022-10-03
test: use separate ports in tests

TimoGlastra:

### DIFF
--- a/packages/rest/jest.config.ts
+++ b/packages/rest/jest.config.ts
@@ -6,7 +6,7 @@ const config: Config.InitialOptions = {
   ...base,
   name: '@aries-framework/rest',
   displayName: '@aries-framework/rest',
-  testTimeout: 12000,
+  testTimeout: 120000,
 }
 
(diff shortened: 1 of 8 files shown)


packages/rest/jest.config.ts
```diff
@@ -6,7 +6,7 @@ const config: Config.InitialOptions = {
   ...base,
   name: '@aries-framework/rest',
   displayName: '@aries-framework/rest',
-  testTimeout: 120000,
+  testTimeout: 12000,
```
janrtvld, line 9: Is this not used from the base config?

janrtvld (approved):
Why the random port changes?

packages/rest/jest.config.ts
TimoGlastra: Ah no this should be changed back

TimoGlastra, line 9:
```suggestion
  testTimeout: 120000,
```

PR #191, opened 2023-02-06
feat: message sorting

KolbyRKunz:
This adds a new hook that consolidates the messages hook so that all messages can be pulled into an application and displayed in a sorted order


TimoGlastra:
What's the status of this PR @KolbyRKunz

KolbyRKunz:
Tested against AFJ 0.3.3 in bifold without any issues. This is was with main merged. @TimoGlastra. Currently do not have a way to test with AFJ 0.4.0.

TimoGlastra:
We can merge for 0.3 now and update to 0.4 later.

Can you remive the package-lock and fix the version?

packages/react-hooks/src/ProofProvider.tsx
```diff
@@ -38,6 +38,11 @@ export const useProofById = (id: string): ProofExchangeRecord | undefined => {
   return proofs.find((p: ProofExchangeRecord) => p.id === id)
 }
 
+export const useProofByConnectionId = (connectionId: string): ProofExchangeRecord[] => {
```
TimoGlastra, line 41: same here (`useProofsByConnectionId`)

packages/react-hooks/src/MessagesProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const useMessages = () => {
+  const messageContext = useContext(MessageContext)
+  if (!MessageContext) {
+    throw new Error('useMessages must be used within a MessageContextProvider')
+  }
+  return messageContext
+}
+
+export const useMessagesByConnectionId = (connectionId: string): BaseRecord[] | undefined => {
```
TimoGlastra, line 22: I assumed that messages would mean basic messages that have been exchanged. Should we call this `useExchanges`, or maybe something else that describes it a bit better?

packages/react-hooks/src/CredentialProvider.tsx
```diff
@@ -38,6 +38,11 @@ export const useCredentialById = (id: string): CredentialExchangeRecord | undefi
   return credentials.find((c: CredentialExchangeRecord) => c.id === id)
 }
 
+export const useCredentialByConnectionId = (connectionId: string): CredentialExchangeRecord[] => {
```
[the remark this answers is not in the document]
KolbyRKunz: I missed the 's' on the end of the method and confused myself as a result.

packages/react-hooks/src/MessagesProvider.tsx
KolbyRKunz: I like that idea and think  it is a good clarification.

PR #175, opened 2022-10-06
feat(react-hooks): support question answer module

seajensen:
This PR adds a react hook to work with the Question-Answer module.

PR #193, opened 2023-02-14
Optionally use the q/a context provider

ryankoch13:
This PR expands on the work from Sabrina here : https://github.com/hyperledger/aries-framework-javascript-ext/pull/175 to leverage the AFJ dependency manager and check if the optional question and answer module is registered prior to using the context provider and react hooks. Have not yet tested it in an agent not importing the question-answer module, but I'm hoping that since the component isn't rendered it doesn't even try do the imports / cause an error


ryankoch13:
Looks like it's now working with apps even if they don't use AFJ question-answer, though the app still needs to list it as a peerDependency. Also, I had to adjust the types and reference the string of the event type directly to avoid import issues, any feedback on how to clean that up would be appreciated.

packages/react-hooks/src/QuestionAnswerProvider.tsx
```diff
@@ -0,0 +1,104 @@
...
+    loading: true,
+  })
+
+  const setInitialState = async () => {
+    if (agent) {
+      if (agent.modules.questionAnswer!) {
+        const questionAnswerMessages = await agent.modules.questionAnswer.getAll()
+        setQuestionAnswerState({ questionAnswerMessages, loading: 'yes-qa' })
+      } else {
+        setQuestionAnswerState({ questionAnswerMessages: [], loading: 'no-qa' })
```
TimoGlastra, line 65: It feels a bit weird to me to have a value of `yes/no-qa` for the `loading` property.

I think we should only load the question answer provider if the question answer module is registered. For now, I think we can just add the quetsion answer module as a dependency and check if it's enabled? It means we do add the dependency, but it's a small pacakge, so it's fine I think.

That way we can do:
1. add an util method that checks whether the question/answer module is registered. We can do this as follows:
  - `Object.values(agent.dependencyManager.registeredModules).find((module) => module instanceof QuestionAnswerModule)`
2. In the `AgentProvider.tsx` we create a value that stores whether the module is enabled:
  - `const isQuestionAnswerModuleEnabled = useMemo(() => isQuestionAnswerModuleEnabled(agent), [agent])`
3. in the render of the `AgentProvider` we only add the question answer provider if the qa module is enabled.
4. If the qa provider is used, we can assume the module is always loaded. Also, we should not assume the module is loaded under `questionAnswer` as you can register the module under any key you wish. Rather, we can inject it:
  - `const questionAnswerApi = agent.dependencyManager.resolve(QuestionAnswerApi)`

packages/react-hooks/src/QuestionAnswerProvider.tsx
```diff
@@ -0,0 +1,104 @@
+import type { Agent } from '@aries-framework/core'
+import type {
+  QuestionAnswerRecord,
+  QuestionAnswerStateChangedEvent,
+  QuestionAnswerEventTypes,
+} from '@aries-framework/question-answer'
```
TimoGlastra, line 6: Adding it as a dev dependency but importing the type, will cause issues if a project doesn't have the `@aries-framework/question-answer` package installed I think. Have you tested this in a proejct without QA installed?

TimoGlastra (changes requested):
I think we need to rework the approach a bit. There's alternatives, so let me know if you don't agree with the solution I proposed

packages/react-hooks/src/QuestionAnswerProvider.tsx
ryankoch13: If we only import the types from the module it doesn't totally error out if the module isn't there, it just doesn't do anything when TypeScript resolves to JavaScript.

packages/react-hooks/src/QuestionAnswerProvider.tsx
ryankoch13: This is a good point, I'm going to check with James and get back to you.

packages/react-hooks/src/recordUtils.ts
```diff
@@ -105,3 +106,15 @@ export const importModule = async (moduleName: string): Promise<any> => {
...
+export const checkModuleEnabled = (agent: Agent, moduleName: any) => {
+  if (!agent) {
+    throw new Error('Agent is required to check if a module is enabled')
+  }
+
+  let enabled = false
+  if (Object.values(agent.dependencyManager.registeredModules).find((module: any) => module instanceof moduleName)) {
+    enabled = true
+  }
+  return enabled
```
TimoGlastra, line 119: Find returns the value when it find something. Although this works, it's not how it's intended to be used

```suggestion
  const foundModule = Object.values(agent.dependencyManager.registeredModules).find((module: any) => module instanceof moduleName))
  
  return foundModule !== undefined
```

TimoGlastra (approved):

JamesKEbert (approved):
LGTM

PR #204, opened 2023-03-29
feat: updated redux-store to AFJ 0.4.x

rapaktech:


packages/redux-store/src/recordListener.ts
```diff
@@ -21,7 +28,7 @@ export const removeRecord = createAction<BaseRecord>('record/remove')
  */
 export const startRecordListeners = (agent: Agent, store: Store) => {
   const onDeleted = (event: RecordDeletedEvent<BaseRecord>) => {
-    store.dispatch(removeRecord(event.payload.record))
+    store.dispatch(removeRecord(event.payload.record as BaseRecord<TagsBase, TagsBase, Record<string, unknown>>))
```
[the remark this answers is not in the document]
rapaktech: @janrtvld @blu3beri @TimoGlastra what do you think?

packages/redux-store/src/utils.ts
```diff
@@ -22,7 +22,7 @@ interface AgentThunkApiConfig {
 
 function createAsyncAgentThunk<Returned, ThunkArg = void>(
   typePrefix: string,
-  payloadCreator: AsyncThunkPayloadCreator<Returned, ThunkArg, AgentThunkApiConfig>
+  payloadCreator: AsyncThunkPayloadCreator<Promise<Returned>, ThunkArg, AgentThunkApiConfig>
```
janrtvld, line 25: How does this suddenly need a promise?

blu3beri: This type is not required. https://github.com/reduxjs/redux-toolkit/blob/c111bba464dd53b4af743b86da53e5de9098ca5a/packages/toolkit/etc/redux-toolkit.api.md?plain=1#L168 wraps the return type in a `MaybePromise` which uses `PromiseLike` which takes care of this. I also tested it locally and it does work all good with `yarn check-types`.

blu3beri (changes requested):

packages/redux-store/src/utils.ts
rapaktech: Yes, it does work. Seems to be a quirk from from my editor's EsLint

PR #207, opened 2023-04-06
chore(redux-store)!: update to afj 0.4.0

morrieinmaas:
Replaces [this PR](https://github.com/hyperledger/aries-framework-javascript-ext/pull/204/files) as Jim has left and I don't have write access to repo.


packages/redux-store/src/recordListener.ts
```diff
@@ -21,7 +26,21 @@ export const removeRecord = createAction<BaseRecord>('record/remove')
  */
 export const startRecordListeners = (agent: Agent, store: Store) => {
   const onDeleted = (event: RecordDeletedEvent<BaseRecord>) => {
-    store.dispatch(removeRecord(event.payload.record))
+    // Extract the record
+    const record = event.payload.record
+    // Retrieve and delete the record if it is of the generic type
+    const retrieveAndDeleteGeneric = async (record: GenRecord) => {
```
morrieinmaas, line 32: @blu3beri not quite sure here (see also [your old comment](https://github.com/hyperledger/aries-framework-javascript-ext/pull/204/files#r1153021830). Might be butchering this. Let me know if you have other things in mind.

morrieinmaas:
Also closes #204

packages/redux-store/package.json
```diff
@@ -29,12 +29,12 @@
     "react-redux": "^7.2.5"
   },
   "devDependencies": {
-    "@aries-framework/core": "^0.2.0",
+    "@aries-framework/core": "^0.4.0-alpha.87",
```
blu3beri, line 32: Would be good to go to the latest for now to make the move to 0.4.0 stable easier.

https://www.npmjs.com/package/@aries-framework/core/v/0.4.0-alpha.104

packages/redux-store/src/recordListener.ts
blu3beri: `genericRecords` would not be able to fetch credentialexchangerecords, proofexchangerecords, etc.

The record might be a type that extends the `BaseRecord` or its `{ id:string, type string }`. 

If it extends a `BaseRecord`, the previous flow was fine. If its an `{id: string, type: string}` it must be fetched manually.

To fetch it maually, although this is getting into some odd dependency injection stuff here so I might get some details wrong, you can resolve the `storageService` from the agent and call `getById`,
https://github.com/hyperledger/aries-framework-javascript/blob/644e860a05f40166e26c497a2e8619c9a38df11d/packages/core/src/storage/StorageService.ts#L62, with the Record Class based on the type. 

@TimoGlastra @genaris I remember something vaguely that this was done to avoid another`getById` call for the `deleteById` method, but if this is the correct way of doing it, it seems like quite a clunky API.

packages/redux-store/src/recordListener.ts
```diff
@@ -8,6 +8,11 @@ import { createAction } from '@reduxjs/toolkit'
 
 import { isRecordType } from './utils'
 
+interface GenRecord {
```
blu3beri, line 11: This can be removed.

packages/redux-store/src/recordListener.ts
blu3beri, line 32: `GenRecord` is again type casted. It can be a full `BaseRecord` which does not need additional fetching. before retrieving any record from storage, it must be checked whether it is a full record, this happens when `delete` is called or if its just the type and id, this happens when `deleteById` is called.

blu3beri (changes requested):
Not the correct solution for the type casting I would say. I think we would either have to do some more complex dependency injection to get to the lower `StorageService` or the `deleteById` event would contain the entire record, unifying the `delete` event from both functions.

packages/redux-store/package.json
morrieinmaas: sure! nice catch. Didn't spot that because I just looked at the unresolved commenst from Jim's PR - my bad

packages/redux-store/src/recordListener.ts
morrieinmaas: aight. could you try rephrase that? I don't get your point.

morrieinmaas: I'm only retrieving it if it's not a full record/has only id and type. SOrry I'm missing your point I think.

blu3beri: `record` in the function interface is now of type `GenRecord` which is just `{id: string, type: string}` which is an unnecessary down-cast of the `BaseRecord` type. Doing this results in having to fetch a record by id for which you already have the record.



```typescript
// pseudo-code

const record: BaseRecord = { ... /* lots of fields */ };

const retrieveAndDeleteRecord = (record: {id: string, type:string }) => {
  // Only call this if the record does not extend `BaseRecord`
  const newRecord = await retrieveRecord(record.id)
  await deleteRecord(newRecord)
}

// if its being called like this, you already have the record ready to be deleted. Fetching it again is unnecessary overhead
retrieveAndDeleteRecord(record)

retrieveAndDeleteRecord({id: 'foo', type: "CredentialExchangeRecord"})

```

morrieinmaas: Thanks!

morrieinmaas:
@blu3beri unsure why this is fialing as it is https://github.com/hyperledger/aries-framework-javascript-ext/actions/runs/4628714463/jobs/8188133138

packages/redux-store/src/recordListener.ts
TimoGlastra: There's not really a need to remove the record here. This event is emitted when the record is already deleted, and the only thing we need to do is to remove it from the redux store.

I think we should just match the API from AFJ and allow either a record, or an object with `type` and `id` to be passed, as that's enough to remove something from the store.

I think we can update the `removeRecord` method to something like this. Haven't tested it, but should work. If it's really difficult. the removeRecord could also just take an id instead of the whole record, but I've made it a record to keep it consistent with the add / update methods.

```
export const removeRecord = <R extends BaseRecordAny>(record: R | { id: string; type: R['type'] }, state: RecordsState<R>): RecordsState<R> => {
```

packages/redux-store/src/recordListener.ts
```diff
@@ -21,7 +21,20 @@ export const removeRecord = createAction<BaseRecord>('record/remove')
...
+      //  No need to fetch it first if we have the id?!
+      await agent.genericRecords.deleteById(record.id)
+    }
+    if (record instanceof BaseRecord) {
+      // BaseRecord
+      store.dispatch(removeRecord(record))
+    } else {
+      // Gen record type
+      void retrieveAndDeleteGenericRecord(record)
+    }
```
TimoGlastra, line 37: This code can stay as is. The deleting is already done. We just need to remove it from the store. As the type of `record` could now also be a simple object with keys `id` and `type`, we jsut need to update the removeRecord method to also support the changes made in AFJ. See my other comment on suggestions on how to resolve

```suggestion
    store.dispatch(removeRecord(event.payload.record))
```

janrxyz:
The `createAsyncAgentThunk` doesn't correctly represent the types of the modules. Is that me or do we need to update that in this pr?

blu3beri (approved):

PR #209, opened 2023-04-26
chore(push-notifications)!: update to afj 0.4.0

morrieinmaas:


morrieinmaas:
@TimoGlastra @blu3beri i've changed the integration test to use 16 and 18 node - is that desired?

packages/push-notifications/src/PushNotificationsModuleConfig.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * PushNotificationsModuleConfigOptions defines the interface for the options of the PushNotificationsModuleConfig class.
+ */
+export interface PushNotificationsModuleConfigOptions {
+  /**
+   * The provider to use - apns or fcm
+   *
+   * @default undefined
+   */
+  provider?: 'fcm' | 'apns'
```
janrxyz, line 10: I tried using the modules, and you can now do this:

`new PushNotificationsFcmModule({ provider: 'apns' })`

which is a bit weird.

blu3beri:
> @TimoGlastra @blu3beri i've changed the integration test to use 16 and 18 node - is that desired?

Yeah that's fine.

packages/push-notifications/samples/sample.ts
```diff
@@ -1,31 +1,27 @@
...
-    name: 'aries push notifications agent',
-    publicDidSeed: '65748374657483920193747564738290',
+  const agent = await setupAgentApns({
+    name: 'aries apns push notifications agent',
   })
 
   // Inject the PushNotificationModule
   // NOTE: This has to be done before initializing the agent
-  const pushNotificationsApnsModule = agent.injectionContainer.resolve(PushNotificationsApnsModule)
+  const pushNotificationsApnsModule = agent.modules.pushNotificationsApns
```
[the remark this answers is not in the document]
morrieinmaas: AH thats because if you go to the `setupAgentApns` in `packages/push-notifications/tests/utils/agent.ts` you'll find that there is no fcm module registered. If you register one on the agent you also get the fcm module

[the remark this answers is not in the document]
morrieinmaas: AH thats because if you go to the setupAgentApns in packages/push-notifications/tests/utils/agent.ts you'll find that there is no fcm module registered. If you register one on the agent you also get the fcm module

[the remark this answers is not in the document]
morrieinmaas: actually there is sth funky going on. i'll fix it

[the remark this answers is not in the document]
morrieinmaas: nice catch - I forgot to register the Api on the Fcm module

morrieinmaas:
@blu3beri how do i get rid of the pending expected integration tests?

packages/push-notifications/src/PushNotificationsModuleConfig.ts
janrxyz: This is still possible.

packages/push-notifications/src/index.ts
```diff
@@ -1,6 +1,7 @@
 import 'reflect-metadata'
 
 export { PushNotificationsApnsModule, PushNotificationsFcmModule } from './modules'
+export { PushNotificationsApnsApi, PushNotificationsFcmApi } from './api'
```
janrxyz, line 4: Is there a reason why you would import the Api and not the Module?

packages/push-notifications/src/PushNotificationsModuleConfig.ts
morrieinmaas: you can put whatever object as options. I put that so that its clear that one can pass options, but they aren't used in the base implementation. This way you can set whatever optionsyou want/need for your own use case (well, one can). Alternatively, options could be removed entirely from the implementation which is also fine

packages/push-notifications/src/index.ts
morrieinmaas: sorry i'm not entirely sure what you mean here

blu3beri:
> @blu3beri how do i get rid of the pending expected integration tests?

I think @TimoGlastra can set new CI pipelines to be required and remove these. I can not do that.

morrieinmaas:
> > @blu3beri how do i get rid of the pending expected integration tests?
> 
> I think @TimoGlastra can set new CI pipelines to be required and remove these. I can not do that.

Ah thanks for clarifying

packages/push-notifications/src/index.ts
morrieinmaas: ah thx for explaining. I see what you mean now. yes, let's remove the export

packages/push-notifications/package.json
```diff
@@ -26,15 +26,18 @@
...
     "@types/jest": "^27.0.2",
+    "@types/node-fetch": "^2.6.3",
+    "indy-sdk": "^1.16.0-dev-1655",
     "ts-node-dev": "^1.1.8",
     "tslog": "^3.2.2",
     "typescript": "~4.4.2"
   },
   "peerDependencies": {
-    "@aries-framework/core": "^0.2.0"
+    "@aries-framework/core": "0.4.0"
```
blu3beri, line 40:
```suggestion
    "@aries-framework/core": "^0.4.0"
```

or is there an unknown quirk I am not familiar with of peerDeps with `^`?

blu3beri (approved):

PR #211, opened 2023-05-17
fix(react-hooks): filtering untyped connections

ryankoch13:
Fixes an issue where the "excluded types" filter will filter connections with no type explicitly specified if it is used.


JamesKEbert (approved):
LGTM

PR #212, opened 2023-06-28
chore: update dev dependencies

TimoGlastra:


packages/react-hooks/src/recordUtils.ts
```diff
@@ -102,18 +101,14 @@ export const recordsRemovedByType = <R extends BaseRecordAny>(
...
 }
 
-export const importModule = async (moduleName: string): Promise<any> => {
-  const importedModule = await import(moduleName)
-  return importedModule
-}
-
-export const checkModuleEnabled = (agent: Agent, moduleName: any) => {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const checkModuleEnabled = (agent: Agent, ModuleClass: any) => {
```
janrxyz, line 105: Any way to loose this `any`?

janrxyz (approved):

packages/react-hooks/src/recordUtils.ts
TimoGlastra: Yes

PR #216, opened 2023-06-29
chore(react-hooks)!: update to afj 0.4.0

TimoGlastra:

BREAKING CHANGE: The package has been updated to work with AFJ 0.4.0. In addition:
- the `agent` prop is now required, and you must create and initialize the agent before passing it to the `AgentProvider`
- The  deprecated `useConnectionByState` and `useConnectionNotInState` methods have been removed. You can use the `useConnections` filters to achieve similar behaviour.


packages/react-hooks/src/recordUtils.ts
```diff
@@ -9,10 +9,11 @@ import type {
 import type { Constructor } from '@aries-framework/core/build/utils/mixins'
 
 import { RepositoryEventTypes } from '@aries-framework/core'
+import { useMemo } from 'react'
 import { map, filter, pipe } from 'rxjs'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-type BaseRecordAny = BaseRecord<any, any, any>
+export type BaseRecordAny = BaseRecord<any, any, any>
```
janrxyz, line 16: Could we maybe add a clarifying comment on what this `BaseRecordAny` is and what it does?

janrxyz (approved):

packages/react-hooks/src/recordUtils.ts
TimoGlastra: Yes!

PR #222, opened 2023-07-28
feat(rest)!: update AFJ to 0.4.0

### LINKED ISSUE #215: Upgrade rest package to use latest aries deps
We are interested in upgrading the rest package to use the latest available versions of `aries-framework/core` and `aries-framework/node`. Our primary driver for this is because as far as we can tell earlier versions do not support the abstraction of hyperledger indy.  We believe that it is important to display a variety of anoncreds registries and would benefit other users of this REST API.

We have read your contributing guidelines and will conform to the standards set out within.

Can you please indicate if you would like us to proceed with this work or if you have other plans as to why we should not proceed?

mattdean-digicatapult:
resolves #215 

Also switches the rest package to use `indy-vdr`, `askar` and `anoncreds` package set removeing `indy-sdk`

So this is a fairly opinionated upgrade with a fair number of breaking changes against the original rest API. The majority of the potentially controversial changes are to the `SchemaController` and the `CredentialDefinitionController` components as the error handling methodology in the `anoncreds` module is very different to that in `indy-sdk`. There are also some pretty substantial changes to the `ProofController`.

I'm now at the point though where I could use an experienced set of eyes to determine if what I've done is sensible. This isn't quite ready but I feel it's close. Any help/advice from existing maintainers would be greatly appreciated. 

A few todo items still remain to be resolved notably:

- [x] review configuration and defaults. Some defaults no longer seem to be valid and some configuration items may have been missed. 
- [x] out-of-band proof API seems to have been completely removed from the framework. My guess looking at the PRs there is that it's somehow been roled into the `oob` functionality but it's unclear how to replicate this or if it's even possible
- [x] on the proof request restrictions there are properties that use a pattern matched key such as `attr::${string}::value` that I cannot reproduce using TSOA. I'm not sure how or if these should be represented.
- [x] test coverage needs improvement in a few areas and I'd like to do that as part of this PR to make sure I have messed anything up. One place I've noticed and marked is in the credential definition creation tests

Next, whilst the tests are currently passing 2 suites fail due to a cleanup issue that still needs to be resolved. Essentially when two of the tests suites shut down I'm getting an error I'm struggling to debug around the Askar wallet being closed:

```
 WalletError: Error closing wallet': Invalid resource handle

      112 |
      113 |   afterAll(async () => {
    > 114 |     await aliceAgent.shutdown()
          |     ^
      115 |     await aliceAgent.wallet.delete()
      116 |     await bobAgent.shutdown()
      117 |     await bobAgent.wallet.delete()

      at AskarWallet.close (node_modules/@aries-framework/askar/src/wallet/AskarWallet.ts:410:13)
      at WalletApi.close (node_modules/@aries-framework/core/src/wallet/WalletApi.ts:97:5)
      at Agent.shutdown (node_modules/@aries-framework/core/src/agent/Agent.ts:217:7)
      at Object.<anonymous> (tests/webhook.test.ts:114:5)
```

Honestly the entire time I've worked on this the tests have felt a bit fragile, partly because of concurrent testing causing collisions in open wallets. I've done my best to fix some of these issues, but the above failure seems intermittent so I'm not convinced I was entirely successful.

FInally I should point out that up to now I've been relying almost exclusively on the tests here to determine if things are working. I don't know if there's any other things I should be checking, but if there is some guidance would be very helpful.


mattdean-digicatapult:
Thanks for the pointer @genaris. I'll be honest I hadn't followed the docs precisely because this package is marked as supporting node 12+. Having tried what you suggested locally is helps ... somewhat. The tests now pass most of the time, but then occationally the same error will occur. Honestly my immediate thought is there's some sort of race condition occuring under the hood that prevents shutdown from always succeeding. Looking at the logs, when the error, occurs there does seem to be some didcom activity taking place after the shutdown is caused and this is then somehow causing the test failure. This is pretty much a gues at this stage.

One question; is the change in https://aries.js.org/guides/0.4/getting-started/set-up/aries-askar something you're happy for me to make? It would mean changing thetop level package in this repository to only support node 18+.

genaris:
> One question; is the change in https://aries.js.org/guides/0.4/getting-started/set-up/aries-askar something you're happy for me to make? It would mean changing thetop level package in this repository to only support node 18+.

IMHO it will not be a problem to do so, given that we are already planning to drop support for node 16 in [AFJ main repository](https://github.com/hyperledger/aries-framework-javascript/issues/1510). We are also discussing to update dependencies in `anoncreds-rs`, `aries-askar` and `indy-vdr` so this is not even needed (see https://github.com/hyperledger/anoncreds-rs/pull/231).

packages/rest/src/cliAgent.ts
```diff
@@ -61,7 +93,37 @@ export async function runRestAgent(restConfig: AriesRestConfig) {
...
+    modules: {
+      connections: new ConnectionsModule({
+        autoAcceptConnections,
+      }),
+      proofs: new ProofsModule({
+        autoAcceptProofs,
+      }),
+      credentials: new CredentialsModule({
+        autoAcceptCredentials,
+      }),
```
genaris, line 108: As you may have noticed, Proofs and Credentials modules support by default only V2 protocols. And they don't set any specific credential/proof format, meaning that they are not actually usable.

I think it will be better to create a configuration where the agent supports both V1 and V2 protocols for each, using legacy Indy credentials and AnonCreds. Look at [`getAskarAnonCredsIndyModules`](https://github.com/hyperledger/aries-framework-javascript/blob/4856b8f6632a4d10faeb1b156397226b64e3b8b6/demo/src/BaseAgent.ts#L110) from AFJ demo for a good example about what I'm referring to.

packages/rest/src/controllers/proofs/ProofController.ts
```diff
@@ -91,21 +150,29 @@ export class ProofController extends Controller {
...
-    const { attributes, predicates, connectionId, ...proposalOptions } = proposal
+    const { attributes, predicates, connectionId, comment } = proposal
 
     try {
-      const presentationPreview = JsonTransformer.fromJSON({ attributes, predicates }, PresentationPreview)
-
-      const proof = await this.agent.proofs.proposeProof(connectionId, presentationPreview, proposalOptions)
+      const proof = await this.agent.proofs.proposeProof({
+        connectionId,
+        protocolVersion: 'v2',
```
genaris, line 167: Here we are hard-coding v2 because this Agent setup supports only that protocol. I think `RequestProposalOptions` could be extended to accept other parameters, like `protocolVersion` and format (indy / anoncreds).

packages/rest/src/controllers/proofs/ProofController.ts
```diff
@@ -124,22 +191,31 @@ export class ProofController extends Controller {
...
       comment?: string
     },
     @Res() notFoundError: TsoaResponse<404, { reason: string }>,
     @Res() internalServerError: TsoaResponse<500, { message: string }>
   ) {
     try {
-      const proof = await this.agent.proofs.acceptProposal(proofRecordId, proposal)
+      const proof = await this.agent.proofs.acceptProposal({
+        proofRecordId,
+        proofFormats: {
```
genaris, line 211: Here. depending on the format it will use indy or anoncreds

packages/rest/src/controllers/proofs/ProofController.ts
```diff
@@ -152,46 +228,76 @@ export class ProofController extends Controller {
...
     @Res() internalServerError: TsoaResponse<500, { message: string }>
   ) {
-    const { connectionId, proofRequestOptions, ...config } = request
+    const { connectionId, proofRequestOptions } = request
 
     try {
-      const proof = await this.agent.proofs.requestProof(connectionId, proofRequestOptions, config)
+      const proof = await this.agent.proofs.requestProof({
+        connectionId,
+        protocolVersion: 'v2',
```
genaris, line 274: Same as above regarding protocolVersions

packages/rest/src/utils/agent.ts
```diff
@@ -1,80 +1,101 @@
...
+    dependencies: agentDependencies,
+    modules: {
+      connections: new ConnectionsModule({
+        autoAcceptConnections: true,
+      }),
+      proofs: new ProofsModule({
+        autoAcceptProofs: AutoAcceptProof.ContentApproved,
+      }),
+      credentials: new CredentialsModule({
+        autoAcceptCredentials: AutoAcceptCredential.ContentApproved,
```
genaris, line 64: Same comments about module config regarding this default agent

genaris (changes requested):
Amazing job @mattdean-digicatapult ! Very clean and useful. I left some comments regarding agent configuration to support all features we previously did (in terms of protocols and credential formats). There will be room for improvements to make REST Agent more configurable (for instance adding other VDRs than Indy or credential formats than AnonCreds), but I think we can start with this update and then think about that.

An additional point that would make this update match the features from previous release is DID creation and import: as you've seen, we don't have a public DID setting anymore, so we'll need to either register or import an existing DID in order to be ables to sign Indy ledger write transactions. So my suggestion would be to update DidController to have an interface like the following:

- `/dids/resolve/{did}`: resolve a DID (same as current `/dids/{did}`)
- `/dids/create`: create, register and store a new DID (take parameters from `agent.dids.create`)
- `/dids/import`: import an existing DID (this is the typical case that replaces current `publicDidSeed` functionality)

There could be some other methods (like `update`) but I think this would be enough and quite straightforward to implement.

Pretty happy to see this PR. Hopefully we can merge it soon! I promise to give you more feedback quickly than before 😄

packages/rest/src/cliAgent.ts
TimoGlastra: > As you may have noticed, Proofs and Credentials modules support by default only V2 protocols. And they don't set any specific credential/proof format, meaning that they are not actually usable.

We should change this...

TimoGlastra:
This is great @mattdean-digicatapult!! Agree with @genaris in extending the interface to allow for more options. In general I think we should keep the interface of the REST api calls as close as possible to the Api classes methods and parameters, so there is consistency in features, options, and how you use the apis.

So e.g. if a call to createOffer is an object, with `connectionId`, `proofFormats` and `protocolVersion`, the REST api interface should be exactly the same. This way changes to the api in AFJ would not mean we have to reinvent the API, as we jsut follow the existing API


> out-of-band proof API seems to have been completely removed from the framework. My guess looking at the PRs there is that it's somehow been roled into the oob functionality but it's unclear how to replicate this or if it's even possible

You are correct that this is now handled by the oob functionalnity. Basically what is possible now:
1. Create credential offer or create proof request
2. `createLegacyConnectionlessInvitation` or `createInvitation`  on the oob module. In both cases you pass the message to the method.

> review configuration and defaults. Some defaults no longer seem to be valid and some configuration items may have been missed.

packages/rest/src/controllers/proofs/ProofController.ts
```diff
@@ -1,18 +1,77 @@
...
-import { Agent, JsonTransformer, PresentationPreview, RecordNotFoundError } from '@aries-framework/core'
-import { JsonEncoder } from '@aries-framework/core/build/utils/JsonEncoder'
+import { Agent, RecordNotFoundError } from '@aries-framework/core'
 import { Body, Controller, Delete, Example, Get, Path, Post, Query, Res, Route, Tags, TsoaResponse } from 'tsoa'
 import { injectable } from 'tsyringe'
 
 import { ProofRecordExample, RecordId } from '../examples'
 import { RequestProofOptions, RequestProofProposalOptions } from '../types'
 
+const maybeMapValues = <V, U>(
```
TimoGlastra, line 13: Would be good to move this to an utils directory

packages/rest/src/controllers/proofs/ProofController.ts
```diff
@@ -152,46 +228,76 @@ export class ProofController extends Controller {
...
-    const proof = await this.agent.proofs.createOutOfBandRequest(proofRequestOptions, requestOptions)
-
-    return {
-      proofUrl: `${this.agent.config.endpoints[0]}/?d_m=${JsonEncoder.toBase64URL(
-        proof.requestMessage.toJSON({ useLegacyDidSovPrefix: this.agent.config.useLegacyDidSovPrefix })
-      )}`,
-      proofRecord: proof.proofRecord,
-    }
-  }
+  // TODO: Represent out-of-band proof
```
TimoGlastra, line 231: This can be just `create-request` and then return an object with the proofRecord and the requestMessage (follow API from `createRequest` on proofs module. Then the user can call the oob api with the message

packages/rest/src/controllers/types.ts
```diff
@@ -44,14 +41,14 @@ export interface DidResolutionResultProps {
...
 export interface ProofRequestMessageResponse {
   message: string
-  proofRecord: ProofRecord
+  proofRecord: ProofExchangeRecord
 }
 
-type CredentialFormats = [IndyCredentialFormat]
-type CredentialServices = [V1CredentialService, V2CredentialService]
+type CredentialProtocols = [V2CredentialProtocol]
+type CredentialFormats = [CredentialFormat]
```
TimoGlastra, line 48: We probably also want to support v1 protocol + legacy indy  /anoncreds formats

packages/rest/src/controllers/types.ts
```diff
@@ -175,20 +172,55 @@ export interface ConnectionInvitationSchema {
   imageUrl?: string
 }
 
-export interface RequestProofOptions extends ProofRequestConfig {
+export interface RequestProofOptionsProofRequestRestriction {
```
TimoGlastra, line 175: I think we should match this with the API as we have in AFJ, and not create custom properties here.


The API in AFJ is:

```
{
  connectionId: string
  protocolVersion: string // 'v1' or 'v2'
  proofFormats: {
   <dynamic interface>
  }
}
```

You can use the whole `RequestProofOptions` type from AFJ to get the correct types and input interface.


You can define the proof protocols type once like this:

```
type ProofProtocols = [V1ProofProtocol, V2ProofProtocol<[LegacyIndyProofFormatService, AnonCredsProofFormatService]>]
```

and then in the rest api calls you can use as the type:

```
RequestProofOptions<ProofProtocols>
```

One thing you may run into is that the generated swagger is invalid, and this is due to some bugs in tsoa: https://github.com/lukeautry/tsoa/issues?q=is%3Aopen+is%3Aissue+author%3A%40me+sort%3Aupdated-desc

Maybe this is resolved by updating to the latest version, but if still the case, you need to define the top level object yourself, but can still infer the format types: 

See this as an example: https://github.com/hyperledger/aries-framework-javascript-ext/blob/main/packages/rest/src/controllers/types.ts#L125


For the request proof interface this would look something like:
```ts
// Define once, can be reused
type ProofFormats = [LegacyIndyProofFormat, AnonCredsProofFormat]

interface RestRequestProofOptions {
  connectionId: string
  autoAcceptProof?: AutoAcceptProof
  comment?: string
  proofFormats: ProofFormatPayload<ProofFormats, 'createRequest'>
}
```

It may seem a bit complex, but it helps in keeping the API of the rest package consistent with what we have in the JS api.

Let me know if you have any questions

TimoGlastra: In general, looking at how it's done in the credentials controller is a good example, as this one was already updated to match the new api from 0.3.x which overhauled the credentials api but not yet the proofs api

mattdean-digicatapult: I understand the reasonning here but there are some problems with `attr::${string}::marker` and `attr::${string}::value` specifically. Notably tsoa does not deal well with types containing patterned keys like:

```ts
type Foo {
  [key: `attr::${string}::marker`]: '1' | '0'
  [key: `attr::${string}::value`]: string,
}
```

including types like this in a body will result in tsoa throwing an internal `GenerateMetadataError` when we build the spec and routes. (I'll see if I can find the issue that discussed this but I can't find it to hand)

This essentially leaves us with two choices:

1. match the API using just a `{ [key: string]: unknown }` and then parse out and validate the `marker` and `value`entries by hand
2. introduce new properties that represent these concepts more neatly within tsoa

I understand the value of consistency but choosing option 1. will lead to a poorly documented OpenAPI type and potentially buggy and error prone parseing logic being done by hand, which feels messy to say the least.

I don't really like the idea of the API being dictated by an issue with tooling, bu tin this instance I felt it was the lesser evil

Thoughts?

TimoGlastra: I agree with you here. Maybe we can use the structure of the class in this case? So it will look something like:

```
restrictions: [
        {
          attributeValues: {
            test_prop: 'test_value',
            test_prop2: 'test_value2',
          },
          attributeMarkers: {
            test_prop: true,
            test_prop2: true,
          },
        },
      ],
```

mattdean-digicatapult:
I think I've covered the review comments now if somebody would be able to have a look. On the new Did routes in particular there is a bucket tonne of oppinionation

dblane-digicatapult:
@TimoGlastra Was wondering if you had a chance to review Matts subsequent changes?

Thanks

packages/rest/src/utils/agent.ts
```diff
@@ -1,80 +1,163 @@
...
+  modules extends RestAgentModules = {
+    connections: ConnectionsModule
+    proofs: ProofsModule<
+      [V1ProofProtocol, V2ProofProtocol<[LegacyIndyProofFormatService, AnonCredsProofFormatService]>]
+    >
+    credentials: CredentialsModule<[V1CredentialProtocol, V2CredentialProtocol]>
+    anoncreds: AnonCredsModule
+  }
+> = Agent<modules>
+
```
berendsliedrecht, line 55: Can we not infer these types from the modules? If not, this is perfectly fine but if we can do something like `type RestAgent = Agent<ReturnType<getAgentModules>>` (I think) that would be better.

packages/rest/src/utils/helpers.ts
```diff
@@ -1,6 +1,19 @@
 import { JsonTransformer } from '@aries-framework/core'
 import { JsonEncoder } from '@aries-framework/core/build/utils/JsonEncoder'
 
+export function maybeMapValues<V, U>(
+  transform: (input: V) => U,
+  obj?: {
+    [key: string]: V
+  }
```
berendsliedrecht, line 8:
```suggestion
  obj?: Record<string, V>
```

berendsliedrecht: Might also be good to define the return type as `Record<string, U> | undefined`

packages/rest/tests/credential.test.ts
```diff
@@ -194,7 +190,7 @@ describe('CredentialController', () => {
             ],
           },
         },
-        autoAcceptCredential: 'always',
+        autoAcceptCredential: 'always' as AutoAcceptCredential,
```
berendsliedrecht, line 193: enum can be used here.

berendsliedrecht (commented):
Awesome work! Some small, non-blocking, issues but overall looks good!

mattdean-digicatapult:
Apologies for the delay in finishing this off. I've been pulled onto other projects but will hopefully find some time in the next couple of weeks

TimoGlastra:
Superseded by #256

PR #256, opened 2024-03-27
feat(rest)!: upgrade to credo 0.5

TimoGlastra:
Updates the rest API wrapper to Credo 0.5, based on #222. It upgrade to use Aries Askar, Indy VDR and AnonCreds RS, and In addition it also adds support for multi-tenancy.

A follow up PR will add documentation, an easy to run docker image, as well as openid4vc support.

Co-authored-by: Matthew Dean <matthew.dean@digicatapult.org.uk>
Signed-off-by: Matthew Dean <matthew.dean@digicatapult.org.uk>


packages/rest/src/controllers/didcomm/basic-messages/BasicMessagesController.ts
```diff
@@ -0,0 +1,72 @@
...
+@injectable()
+export class DidCommBasicMessagesController extends Controller {
+  /**
+   * Retrieve basic messages by connection id
+   *
+   * @param connectionId Connection identifier
+   * @returns BasicMessageRecord[]
+   */
+  @Example<DidCommBasicMessagesRecord[]>([basicMessageRecordExample])
+  @Get('/')
```
berendsliedrecht, line 28: Oh thats nice! Good to just use query params as the actual basic message query, hadn't thought about that :).

packages/rest/src/controllers/didcomm/basic-messages/BasicMessagesController.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * Send a basic message to a connection
+   *
+   * @param connectionId Connection identifier
+   * @param content The content of the message
+   * @returns BasicMessageRecord
+   */
+  @Example<DidCommBasicMessagesRecord>(basicMessageRecordExample)
+  @Post('/')
```
berendsliedrecht, line 53: Time will tell, but maybe a `post` on root can be a bit unclear that it will actually send a message. A `/send` endpoint (or `send-message` to align with module API) might be more clear.

I am fine with either, just worth writing a note about.

packages/rest/src/controllers/didcomm/connections/ConnectionsController.ts
```diff
@@ -0,0 +1,137 @@
...
+    return connections.map(connectionRecordToApiModel)
+  }
+
+  /**
+   * Retrieve connection record by connection id
+   * @param connectionId Connection identifier
+   * @returns ConnectionRecord
+   */
+  @Example<DidCommConnectionsRecord>(connectionsRecordExample)
+  @Get('/:connectionId')
```
berendsliedrecht, line 56: Can't we just reuse the query for this? API call will be slightly larger but its a nice a consistent way to fetch by a query.

packages/rest/src/controllers/didcomm/connections/ConnectionsController.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    return connectionRecordToApiModel(connection)
+  }
+
+  /**
+   * Deletes a connection record from the connection repository.
+   *
+   * @param connectionId Connection identifier
+   */
+  @Delete('/:connectionId')
```
berendsliedrecht, line 73: Ah I commented too early... I guess with the Delete method it makes sense to separate it to stay conistent method-wise.

packages/rest/src/controllers/didcomm/connections/ConnectionsController.ts
```diff
@@ -0,0 +1,137 @@
...
+   * Accept a connection response as invitee by sending a trust ping message
+   * for the connection with the specified connection id.
+   *
+   * This is not needed when auto accepting of connection is enabled.
+   *
+   * @param connectionId Connection identifier
+   * @returns ConnectionRecord
+   */
+  @Example<DidCommConnectionsRecord>(connectionsRecordExample)
+  @Post('/:connectionId/accept-response')
```
berendsliedrecht, line 122: Is there a specific convention followed? Normally I go for `/accept-response/:connection-id` but I have no idea which is more common.

packages/rest/src/controllers/didcomm/connections/ConnectionsControllerTypes.ts
```diff
@@ -0,0 +1,58 @@
...
+  protocol?: HandshakeProtocol
+  outOfBandId?: string
+  invitationDid?: Did
+  connectionTypes?: Array<ConnectionType | string>
+  previousDids?: Array<Did>
+  previousTheirDids?: Array<Did>
+}
+
+export function connectionRecordToApiModel(record: CredoConnectionRecord): DidCommConnectionsRecord {
+  return {
```
berendsliedrecht, line 32: We can't use the record.toJSON() for this?

packages/rest/src/controllers/didcomm/credentials/CredentialsController.ts
```diff
@@ -0,0 +1,281 @@
...
+
+  /**
+   * Initiate a new credential exchange as holder by sending a propose credential message
+   * to the connection with a specified connection id.
+   *
+   * @param options
+   * @returns CredentialExchangeRecord
+   */
+  @Example<DidCommCredentialsExchangeRecord>(credentialExchangeRecordExample)
+  @Post('/propose-credential')
```
berendsliedrecht, line 112: Would it make sense to have the connection id as a path param? 

Again, I have no major preference to either API, but could be worth considering if it makes it more clear.

berendsliedrecht: However I do really like the fact to just pass the options along to credo and have the logic there. maintenance is a lot easier with this implementation.

packages/rest/src/controllers/didcomm/out-of-band/OutOfBandController.ts
```diff
@@ -0,0 +1,266 @@
...
+import type { PlaintextMessage } from '@credo-ts/core/build/types'
+
+import {
+  AgentMessage,
+  JsonTransformer,
+  OutOfBandInvitation,
+  RecordNotFoundError,
+  ConnectionInvitationMessage,
+} from '@credo-ts/core'
+import { parseMessageType, supportsIncomingMessageType } from '@credo-ts/core/build/utils/messageType'
```
berendsliedrecht, line 11: Should we export these or is this quite niche? Same as with the PlaintextMessage type?

packages/rest/src/controllers/didcomm/out-of-band/OutOfBandController.ts
```diff
@@ -0,0 +1,266 @@
...
+@Tags('DIDComm Out Of Band')
+@Route('/didcomm/out-of-band')
+@Security('tenants', ['tenant'])
+@injectable()
+export class OutOfBandController extends Controller {
+  /**
+   * Retrieve all out of band records by query
+   */
+  @Example([outOfBandRecordExample])
+  @Get()
```
berendsliedrecht, line 44: Ah the `'/'` is not required, cool!

packages/rest/src/controllers/didcomm/out-of-band/OutOfBandController.ts
```diff
@@ -0,0 +1,266 @@
...
+
+    if (!outOfBandRecord) {
+      this.setStatus(404)
+      return apiErrorResponse(`Out of band record with id "${outOfBandId}" not found.`)
+    }
+    return outOfBandRecordToApiModel(outOfBandRecord)
+  }
+
+  /**
+   * Creates an outbound out-of-band record containing out-of-band invitation message defined in
```
berendsliedrecht, line 75: Does TSOA pick this comment up and display it in swagger? Would be cool!

packages/rest/src/controllers/didcomm/out-of-band/OutOfBandController.ts
```diff
@@ -0,0 +1,266 @@
...
+   * Aries RFC 0434: Out-of-Band Protocol 1.1.
+   */
+  @Example({
```
berendsliedrecht, line 78: Shouldn't this be in the Examples file>?

packages/rest/src/controllers/tenants/TenantsController.ts
```diff
@@ -0,0 +1,30 @@
...
+import { Body, Controller, Post, Request, Route, Security, Tags } from 'tsoa'
+import { injectable } from 'tsyringe'
+
+import { RequestWithRootTenantAgent } from '../../authentication'
+
+import { tenantRecordToApiModel } from './TenantsControllerTypes'
+
+@Tags('Tenants')
+@Route('/tenants')
+@Security('tenants', ['admin'])
```
berendsliedrecht, line 13: what is the difference bwteen admin and default?

packages/rest/src/middlewareTemplate.ts.hbs
```diff
@@ -0,0 +1,201 @@
+/* tslint:disable */
```
berendsliedrecht, line 1: Do we need to push this?

packages/rest/src/routes/swagger.json
```diff
@@ -1,111 +1,110 @@
 {
```
berendsliedrecht, line 1: Would be possible to have all these generated files inside a `generated` folder? Just to keep it clear what is generated by computers and humans :).

packages/rest/src/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface ApiError {
```
berendsliedrecht, line 1: Can this be moved to something related to errors? Or maybe rename the file? `types.ts` is quite vague.

packages/rest/src/utils/agent.ts
```diff
@@ -1,79 +1,196 @@
...
+      ],
+    }),
+  } as const
+
+  const modules: typeof baseModules & { tenants?: TenantsModule<typeof baseModules>; indyVdr?: IndyVdrModule } =
+    baseModules
+
+  if (options.multiTenant) {
+    modules.tenants = new TenantsModule({
+      sessionLimit: Infinity,
```
berendsliedrecht, line 126: Thats very scalable! wonder what happens when you exceed f32::MAX sessions :').

packages/rest/src/utils/agent.ts
```diff
@@ -1,79 +1,196 @@
...
   endpoints: string[]
-  port: number
-}) => {
-  const logger = new TsLogger(LogLevel.debug)
+  extraAnonCredsRegistries?: AnonCredsRegistry[]
+  httpInboundTransportPort?: number
+  multiTenant?: boolean
+}): Promise<RestRootAgent | RestRootAgentWithTenants> => {
+  // FIXME: logger should not be enabled by default
+  const logger = new TsLogger(LogLevel.off)
```
berendsliedrecht, line 154: Would this be possible to set with the start script? with something like `-vvv` for loglevel=3?

packages/rest/src/utils/response.ts
```diff
@@ -0,0 +1,16 @@
+import type { ApiError } from '../types'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function alternativeResponse<T>(input: T): any {
```
berendsliedrecht, line 4: `T` can be return type as well right?

packages/rest/tsconfig.build.json
```diff
@@ -2,7 +2,9 @@
   "extends": "../../tsconfig.build.json",
   "compilerOptions": {
     "outDir": "./build",
-    "types": []
+    "types": [],
```
berendsliedrecht, line 5: We can have node types here right?

packages/rest/tsconfig.build.json
```diff
@@ -2,7 +2,9 @@
...
+    "skipLibCheck": true,
```
berendsliedrecht, line 6: Thats already set in the abse tsconfig right?

berendsliedrecht (approved):
Nothing blocking at all! Really nice changes. Looks like keeping the API up-to-date and adding new features was quite a lot for this PR, but with the current structure it seems quite easy to do, nice!

packages/rest/README.md
```diff
@@ -32,153 +32,8 @@ The Credo REST API is the most convenient way for self-sovereign identity (SSI)
 
 - ⭐ **Endpoints** to create connections, issue credentials, and request proofs.
 - 💻 **CLI** that makes it super easy to start an instance of the REST API.
-- 🌐 **Interoperable** with all major Aries implementations.
+- 🌐 **Interoperable** with all major Aries implementations and support for OpenID4VC.
```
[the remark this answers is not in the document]
TimoGlastra: yes,  can remove it for now

packages/rest/src/controllers/agent/AgentController.ts
```diff
@@ -1,30 +1,29 @@
...
-  }
-
   /**
    * Retrieve basic agent information
    */
   @Get('/')
-  public async getAgentInfo(): Promise<AgentInfo> {
+  @Example(agentInfoExample)
+  public async getAgentInfo(@Request() request: RequestWithRootAgent): Promise<AgentInfo> {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
TimoGlastra: Yes we want to strip some properties from the config . ADded a comment

packages/rest/src/controllers/anoncreds/__tests__/AnonCredsController.test.ts
```diff
@@ -0,0 +1,196 @@
...
+  })
+
+  afterAll(async () => {
+    await agent.shutdown()
+    await agent.wallet.delete()
+  })
+
+  describe('get schema by id', () => {
+    test('should return schema', async () => {
+      // FIXME: we need to encode the schema-id to make it work ....
```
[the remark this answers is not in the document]
TimoGlastra: No just good to know. If you paste it in, it will give a 404 error as it addes /slashes and thus actually accesses a different route.

packages/rest/src/controllers/didcomm/basic-messages/BasicMessagesController.ts
TimoGlastra: Yes makes sense. There's quite some endpoints I think that we need to change in this case

packages/rest/src/controllers/didcomm/connections/ConnectionsController.ts
TimoGlastra: Yes it's nice to have a get by Id I think, and a get multiple by query

packages/rest/src/controllers/didcomm/connections/ConnectionsController.ts
TimoGlastra: this is restful practice. You do an operation on a connection. Otherwise you'd acces "AcceptResponse" resource with connection id

Format is always "<resource-category>/<resource-id>/<resource-specific-action>"

E.g. `books/12/delete` = From the books category, select the resource with id 12 and then perform the delete operation on that resource

packages/rest/src/controllers/didcomm/connections/ConnectionsControllerTypes.ts
TimoGlastra: Because it will include tags / metadata and some other fields and it messed with the types. So I'd rather have a simple facade that makes sure the Swagger Type makes sense and isn't bloated with internal storage propertiees.

So mainly preference on my side to be explicity because all the tsoa magic and other things add a lot of weird behaviour

packages/rest/src/controllers/didcomm/credentials/CredentialsController.ts
TimoGlastra: This is again restful best practice. We're on the `/credentials` endoint here. So passing a connectionId doesn't make sense (for the other endpoints where a credentialExchangeId is expected it does inlcude it in the path).

If we would want to pass a connectionId in path, we should add the endpoint under the connection resource:
`/didcomm/connections/<connectionId>/propose-credential`

However that gets messy as you can also do some credential operations without connectionId. So connectionId is an input parameter to create a credential exchange resource.

packages/rest/src/controllers/didcomm/out-of-band/OutOfBandController.ts
TimoGlastra: Yes we should probably export these

packages/rest/src/controllers/didcomm/out-of-band/OutOfBandController.ts
TimoGlastra: Yes it does :)

packages/rest/src/controllers/didcomm/out-of-band/OutOfBandController.ts
TimoGlastra: yes :)

packages/rest/src/controllers/tenants/TenantsController.ts
TimoGlastra: Admin can only be accessed as tenant admin, which is only availalbe when the tenant module is active (--multi-tenant enabled). So it means you can't call these endpoints if tenants module is not enabled.

packages/rest/src/middlewareTemplate.ts.hbs
TimoGlastra: Yes, we need a custom template because it doesn't call `next()` at the end by default, and we need to have "afterware" to be able to close a tenant session. 

Spent a lot of time on this, and this is the only way I could find, sadly

packages/rest/src/routes/swagger.json
TimoGlastra: Sure 👍

packages/rest/src/utils/agent.ts
TimoGlastra: Yeah the session implementation is a bit broken in Credo, I want to address it for 0.6, as it counts very weird so it's not a good indication of how much resources you're using

packages/rest/src/utils/agent.ts
TimoGlastra: Yes we can do it for the CLI part already --log-level. I need to improve the non-cli (import rest server) api a bit. Will look at it in my next PR if okay

packages/rest/src/utils/response.ts
TimoGlastra: No as then the return type on the method is wrong. This is my workaround to have alternative responses in TSOA. The proper way to do it (inject error handler) doesn't allow you to set examples for alternative status codes.

This way the OpenAPI / swagger is nice, which is most important to me

Issue #275, opened 2024-06-22
[rest] disable SingleContextStorageLruCache when used in multi-tenant mode


amanji:
While probably an edge case, this prevents a tenant from being able to issue credentials to another tenant in the same agent instance. Or perhaps worded another way it prevents a tenant from caching a credential definition that has already been previously cached by another tenant. Not sure if there may be scenarios where this is applicable.

TimoGlastra:
Makes sense! I think we can enable the in-memory cache by default

PR #277, opened 2024-06-26
fix: enable in-memory cache when in multi-tenant mode

amanji:
Resolves: #275


TimoGlastra (approved):

TimoGlastra:
I think we need to update shared components / node versions to get CI working again

amanji:
I regenerated the yarn lockfile but now `@credo-ts/cheqd` can't be resolved in the tests.

amanji:
@TimoGlastra is there anything specific I need to get the tests to pass here? I'm a bit stuck on why they may be failing.

TimoGlastra:
There have been some changes to the OpenId4VC module which is causing these issues in tests I think

